PR #2, opened 2022-10-18
tests: initial setup for Candlepin-based tests

ptoscano:
Integration tests for the rhc role will definitely need a Candlepin server for registration.

As a way to easily test scenarions, add an helper playbook that either deploys Candlepin using the upstream Ansible role, or assumes an external Candlepin server is available anywhere. In the latter case, the `LSR_RHC_TEST_DATA` environment variable will point to a YAML file with variables specifying a `lsr_rhc_test_data` dictionary with candlepin-related variables, e.g.:

```yaml
lsr_rhc_test_data:
  candlepin_host: candlepin.server.hostname
  candlepin_port: port
  candlepin_prefix: /candlepin
  candlepin_insecure: 1
```

The approach is taken from the current way the Ansible role is used in the Cockpit CI to generate an image with Candlepin installed as a container.

Add a very simple tests_default test that checks the role works fine without registering.

Signed-off-by: Pino Toscano <ptoscano@redhat.com>


tests/roles/linux-system-roles.rhc
```diff
@@ -0,0 +1 @@
+../..
```
richm, line 1: making `linux-system-roles.rhc` the symlink will cause circular references - instead, create the directory `tests/roles/linux-system-roles.rhc` and in this directory create symlinks to `../../../tasks`, `../../../defaults`, etc.

tests/tasks/setup_candlepin.yml
```diff
@@ -0,0 +1,89 @@
...
+        src: "/tmp/candlepin-role-repo/"
+        dest: "{{ playbook_dir }}/roles/candlepin"
+        state: link
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Create empty/fake rvm.ruby module
+      file:
+        dest: "{{ playbook_dir }}/roles/rvm.ruby"
+        state: directory
```
richm, line 36:
```suggestion
        state: directory
        owner: root
        group: root
        mode: "0777"
```
to satisfy ansible-lint error

tests/tasks/setup_candlepin.yml
```diff
@@ -0,0 +1,89 @@
...
+
+- name: Deploy Candlepin
+  block:
+    - name: Clone ansible-role-candlepin repo
+      git:
+        repo: https://github.com/candlepin/ansible-role-candlepin.git
+        version: master
+        dest: /tmp/candlepin-role-repo
+      delegate_to: 127.0.0.1
+      become: false
```
richm, line 23: This will likely conflict if one or more tests are run at the same time on the same control node (i.e. localhost i.e. 127.0.0.1) which our CI system does.   Instead, use the `tempfile` module to create a per-test temporary working directory: https://docs.ansible.com/ansible/2.9/modules/tempfile_module.html#tempfile-module
```
  block:
    - name: Create working directory for test
      tempfile:
        state: directory
        prefix: lsr_
        suffix: _rhc
      register: __rhc_test_dir
      delegate_to: 127.0.0.1
      become: false

    - name: Clone ansible-role-candlepin repo
      git:
        repo: https://github.com/candlepin/ansible-role-candlepin.git
        version: master
        dest: "{{ __rhc_test_dir.path }}/role-repo"
      delegate_to: 127.0.0.1
      become: false
```

tests/tasks/setup_candlepin.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    - name: Create role symlink
+      file:
+        src: "/tmp/candlepin-role-repo/"
```
richm, line 27:
```suggestion
        src: "{{ __rhc_test_dir.path }}/role-repo"
```

tests/tasks/setup_candlepin.yml
```diff
@@ -0,0 +1,89 @@
...
+        dest: "{{ playbook_dir }}/roles/candlepin"
```
richm, line 28:
```suggestion
        dest: "{{ __rhc_test_dir.path }}/roles/candlepin"
```

tests/tasks/setup_candlepin.yml
```diff
@@ -0,0 +1,89 @@
...
+        state: link
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Create empty/fake rvm.ruby module
+      file:
+        dest: "{{ playbook_dir }}/roles/rvm.ruby"
```
richm, line 35:
```suggestion
        dest: "{{ __rhc_test_dir.path }}/roles/rvm.ruby"
```

tests/tasks/setup_candlepin.yml
```diff
@@ -0,0 +1,89 @@
...
+      become: false
+
+    - name: Install collection requirements of candlepin role
+      command:
+        argv:
+          - ansible-galaxy
+          - collection
+          - install
+          - "-r"
+          - "{{ playbook_dir }}/tasks/setup_candlepin_requirements.yml"
```
richm, line 47:
```suggestion
          - "{{ playbook_dir }}/tasks/setup_candlepin_requirements.yml"
          - "-p"
          - "{{ __rhc_test_dir.path }}/ansible_collections"
```

tests/tasks/setup_candlepin.yml
ptoscano: hm is root:root correct here? this will be created in my own source directory

tests/tasks/setup_candlepin.yml
ptoscano: hm TBH i noticed the same approach as used in the `certificates` role for deploying freeipa; another possibility i thought about was to avoid the cloning in `/tmp` + symlink and directly clone under `tests/roles/` (which i guess has the same issue)

maybe i'm missing something: if i create and use a temporary directory as you shown, how is that used by ansible automatically?

richm: > hm TBH i noticed the same approach as used in the `certificates` role for deploying freeipa; another possibility i thought about was to avoid the cloning in `/tmp` + symlink and directly clone under `tests/roles/` (which i guess has the same issue)

There's only 1 test that does that in the certificate role, so there is no chance of conflict if multiple tests are run in parallel.

> 
> maybe i'm missing something: if i create and use a temporary directory as you shown, how is that used by ansible automatically?

It isn't - that is, if you use `include_role`, I'm not sure how you tell it to look in the temporary directory for the role + collections.  Another way to do this would be to use `command: ansible-playbook ....` to install candlepin, and create a short playbook file in the temp directory.  I need to think about this some more.

tests/tasks/setup_candlepin_requirements.yml
```diff
@@ -0,0 +1,6 @@
+---
+collections:
+  - name: ansible.posix
+  - name: community.general
+  - name: community.mysql
+  - name: community.postgresql
```
richm, line 6: mysql and postgresql?

tests/tasks/setup_candlepin.yml
richm: I think what you have here is ok for now.  Maybe just a note to developers to tell them to clean up their ~/.ansible directory after running the test to remove the collections installed by the test.

richm:
I think the way you are installing the role/collections is ok, at least for now - we'll see if it causes problems in CI.  Just address the couple of minor comments and I think we're good.

tests/tasks/setup_candlepin_requirements.yml
ptoscano: the "candlepin" role has tasks for both; even though we configure to not depploy mariadb, it's easier to pull both those collections than working around in funky ways

tests/tasks/setup_candlepin.yml
ptoscano: > Maybe just a note to developers to tell them to clean up their ~/.ansible directory after running the test to remove the collections installed by the test.

Hm with my latest changes I don't have anything installed locally in my `~/.ansible/` by the tests, everything is in `.tox/`. Do you get anything in your ~?

richm:
[citest]

PR #4, opened 2022-11-02
Basic registration & unregistration

ptoscano:
Add a very simple invocation of the `redhat_subscription` module to register & unregister a system; include a simple test for it.

See details of the changes in each commit.


tasks/subscription-manager.yml
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Call subscription-manager
+  redhat_subscription:
+    state: "{{ rhc_state | d('present') }}"
+    username: "{{ rhc_auth.login.username | d(omit) }}"
+    password: "{{ rhc_auth.login.password | d(omit) }}"
+    activationkey: "{{ rhc_auth.activation_keys.keys | join(',')
+  if ( rhc_auth.activation_keys.keys is defined ) else omit }}"
```
richm, line 9:
```suggestion
      if ( rhc_auth.activation_keys.keys is defined ) else omit }}"
```
suggest to indent

ptoscano: ah ok, will do, thanks (... i shamelessly copied it from other roles :) )

README.md
```diff
@@ -10,18 +10,50 @@ RHEL repositories, and subscription-manager is installed by default.
...
+in case a system may need to connect (e.g. in case it was not before).
+There are few possible authentication methods; only one can be specified at
+a time.
+
+For username/password, authentication, specify the `login` dictionary:
+```yaml
+rhc_auth:
+  login:
+    username: "your-username"
+    password: "your-password"
```
richm, line 30:
```suggestion
    password: !vault |
      $ANSIBLE_VAULT;1.2;AES256;dev
      ....
```

README.md
```diff
@@ -10,18 +10,50 @@ RHEL repositories, and subscription-manager is installed by default.
...
+```
+
+For activation keys authentication, specify the `activation_keys` dictionary:
+```yaml
+rhc_auth:
+  activation_keys:
+    keys: ["key-1", ...]
```
richm, line 37:
```suggestion
    keys:
      - !vault |
          $ANSIBLE_VAULT;1.2;AES256;key1
          ....
      - !vault |
          $ANSIBLE_VAULT;1.2;AES256;key2
          ....
```

README.md
```diff
@@ -10,18 +10,50 @@ RHEL repositories, and subscription-manager is installed by default.
...
+Simple Registration to Red Hat including Insights, assuming authentication
+using username & password:
 
 ```yaml
 - hosts: all
+  vars:
+    rhc_auth:
+      login:
+        username: "your-username"
+        password: "your-password"
```
richm, line 56:
```suggestion
        password: !vault |
          $ANSIBLE_VAULT;1.2;AES256;dev
          ....
```

ptoscano:
Regarding the notes about passwords and activation keys in `README.md` & vault: considering that they are only references/examples on how the role should be used, shouldn't it be up to the users to secure those secrets using a vault or other means? Possibly not the best examples: I noticed also other roles in linux-system-roles have that kind of password keys/parameters documented "as usual strings" in their `README.md`.

richm:
> Regarding the notes about passwords and activation keys in `README.md` & vault: considering that they are only references/examples on how the role should be used, shouldn't it be up to the users to secure those secrets using a vault or other means?

Yes, it is up to the users, but we want to encourage users to use best security practices.  A user new to Ansible might not be aware that Vault exists and secrets should be secured.

> Possibly not the best examples: I noticed also other roles in linux-system-roles have that kind of password keys/parameters documented "as usual strings" in their `README.md`.

Yes, some of the older ones, we did not go back and put in the Vault recommendation and examples.

richm:
I'm getting various failures trying to run the tests on various platforms.
I had to add a task to install git for testing - I guess the candlepin role requires git?
```
diff --git a/tests/tasks/setup_candlepin.yml b/tests/tasks/setup_candlepin.yml
index acacf7a..478a145 100644
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -33,6 +33,11 @@
       delegate_to: 127.0.0.1
       become: false
 
+    - name: Install system requirements of candlepin role
+      package:
+        name: git
+        state: present
+
     - name: Create 'candlepin' user
```

next - the various `rhsm` modules require the subscription-manager package:
```
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: MIT
 ---
+- name: Ensure required packages are installed
+  package:
+    name: "{{ __rhc_packages }}"
+    state: present
+

diff --git a/vars/main.yml b/vars/main.yml
index 347d096..f66baf7 100644
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -1,6 +1,9 @@
 # SPDX-License-Identifier: MIT
 ---
 
+__rhc_packages:
+  - subscription-manager
+
 # ansible_facts required by the role
```
Even with these, I'm getting various errors.  On centos stream 9:
```
TASK [candlepin : Execute deploy script] ***************************************
task path: /home/rmeggins/linux-system-roles/rhc/tests/roles/candlepin/tasks/common/cp_deploy.yml:27
Wednesday 02 November 2022  10:09:02 -0600 (0:00:00.040)       0:00:44.719 **** 
fatal: [/home/rmeggins/.cache/linux-system-roles/centos-9.qcow2]: FAILED! => {
    "changed": false,
    "cmd": [
        "/home/candlepin/devel/candlepin/bin/deployment/deploy",
        "-g",
        "-a",
        "-f",
        "-t"
    ],
    "delta": "0:00:00.064368",
    "end": "2022-11-02 12:09:02.782136",
    "rc": 5,
    "start": "2022-11-02 12:09:02.717768"
}

STDOUT:

Configuring Tomcat 6 found at /var/lib/tomcat6


STDERR:

Redirecting to /bin/systemctl stop tomcat6.service
Failed to stop tomcat6.service: Unit tomcat6.service not loaded.


MSG:

non-zero return code
```
I also tried Fedora, but I'm assuming Fedora is not really a "supported" platform?  Can you register a Fedora system to RHC?

rhel-8 latest hangs at the `candlepin : Execute deploy script` step - something is killing ssh - can't even ssh into the VM.

richm:
looks like this test needs more RAM - I'm trying with a `tests/provision.fmf` like this:
```
standard-inventory-qcow2:
  qemu:
    m: 4096
```

ptoscano:
> Yes, it is up to the users, but we want to encourage users to use best security practices. A user new to Ansible might not be aware that Vault exists and secrets should be secured.

Would it be OK to add a "NB"/disclaimer note in the "rhc_auth" description within "Role Variables"?
This way there is less need to repeat the Vault placeholder.

> I had to add a task to install git for testing - I guess the candlepin role requires git?

Hm looks so, I will add it. Usually you don't notice that because in development setups/environments there is git already installed.

> next - the various `rhsm` modules require the subscription-manager package:

That opens an interesting question (at least IMHO): if I'm registering a new RHEL system, that means there are no configured/enabled repositories where to get content from, so installing things won't work immediately. In RHEL `subscription-manager` is part of the "core" of a system, so it's kind of essential.

Non-RHEL systems such as CentOS and Fedora do have already preconfigured repositories, so installing stuff works also without registration.

That said, `insights-client` (which will be needed for Insights registration) is not installed by default; I think its installation will need to be done after the initial registration.

> Even with these, I'm getting various errors. On centos stream 9:

The "candlepin" role is not yet supported on EL9, so for now testing is limited to EL8.

> I also tried Fedora,

Does it fail too?

> I'm assuming Fedora is not really a "supported" platform?

It shouldn't be difficult to make it work there, however I wouldn't have it as top priority.

> Can you register a Fedora system to RHC?

Sure :)

> rhel-8 latest hangs at the `candlepin : Execute deploy script` step - something is killing ssh - can't even ssh into the VM.
> looks like this test needs more RAM

Hm the internal `rhel-8` image works fine for me; note the aforementioned step takes a while (3-4 minutes).

ptoscano:
> I had to add a task to install git for testing - I guess the candlepin role requires git?

Ah, this should be handled by the role already; reading the other comments you wrote, I think it might be due to the fact that EL9 is not supported yet, so some steps (e.g. installing dependencies, git included) are not performed.

richm:
> > Yes, it is up to the users, but we want to encourage users to use best security practices. A user new to Ansible might not be aware that Vault exists and secrets should be secured.
> 
> Would it be OK to add a "NB"/disclaimer note in the "rhc_auth" description within "Role Variables"? This way there is less need to repeat the Vault placeholder.

Ok.

> 
> > I had to add a task to install git for testing - I guess the candlepin role requires git?
> 
> Hm looks so, I will add it. Usually you don't notice that because in development setups/environments there is git already installed.
> 
> > next - the various `rhsm` modules require the subscription-manager package:
> 
> That opens an interesting question (at least IMHO): if I'm registering a new RHEL system, that means there are no configured/enabled repositories where to get content from, so installing things won't work immediately. In RHEL `subscription-manager` is part of the "core" of a system, so it's kind of essential.

Good point.  I guess a note in the README that either
* `subscription-manager` should be installed as part of system provisioning
* some repo containing `subscription-manager` should be enabled

otherwise there is the bootstrap issue.  *And* the role should do
```
+- name: Ensure required packages are installed
+  package:
+    name: "{{ __rhc_packages }}"
+    state: present
```
where `subscription-manager` is one of the `__rhc_packages` - if it fails, this will be a good reminder to users that the package should already be installed or installable - plus, for testing, we don't have to change the way images are provisioned - in our testing environments, we already ensure that the systems are configured with the necessary repos before running the role.

> 
> Non-RHEL systems such as CentOS and Fedora do have already preconfigured repositories, so installing stuff works also without registration.
> 
> That said, `insights-client` (which will be needed for Insights registration) is not installed by default; I think its installation will need to be done after the initial registration.
> 
> > Even with these, I'm getting various errors. On centos stream 9:
> 
> The "candlepin" role is not yet supported on EL9, so for now testing is limited to EL8.
> 
> > I also tried Fedora,
> 
> Does it fail too?

It failed because I did not install `subscription-manager`

> 
> > I'm assuming Fedora is not really a "supported" platform?
> 
> It shouldn't be difficult to make it work there, however I wouldn't have it as top priority.
> 
> > Can you register a Fedora system to RHC?
> 
> Sure :)
> 
> > rhel-8 latest hangs at the `candlepin : Execute deploy script` step - something is killing ssh - can't even ssh into the VM.
> > looks like this test needs more RAM
> 
> Hm the internal `rhel-8` image works fine for me; note the aforementioned step takes a while (3-4 minutes).

In my case I could not even ssh into the machine - I think it was swapping heavily, which usually means it needs more RAM.

richm:
I made some more changes - now I get to the point where the redhat_subscription module is executed
```
     org_id: "{{ rhc_auth.activation_keys.org | d(omit) }}"
+    server_hostname: "{{ rhc_server.hostname | d(omit) }}"
+    server_port: "{{ rhc_server.port | d(omit) }}"
+    server_prefix: "{{ rhc_server.prefix | d(omit) }}"
+    server_insecure: "{{ rhc_server.insecure | d(omit) }}"


+++ b/tests/tests_register_unregister.yml
@@ -3,7 +3,6 @@
 - name: Basic register/unregister
   hosts: all
   gather_facts: true
-  become: true
   tags:
```
(NOTE: Directives like `become: true` should *not* be put in the playbooks - instead, users should specify these on the command line as needed e.g. `ansible-playbook -b ...`)

However, the `redhat_subscription` module is hung.  I logged into the machine and found the subscription-manager command:
```
/usr/libexec/platform-python /usr/sbin/subscription-manager register --insecure --serverurl 127.0.0.1 --username admin --password admin
```
This is wrong.  It seems there is a mismatch between the `redhat_subscription` module and the `subscription-manager` command on rhel 8.8.  I used this command instead:
```
/usr/sbin/subscription-manager register --insecure --serverurl https://127.0.0.1:8443 --username admin --password admin
```
and that got further, except that it prompted me for the organization:
```
Registering to: 127.0.0.1:8443/candlepin
Hint: User "admin" is member of following organizations: donaldduck, snowwhite, admin
Organization: donaldduck 
The system has been registered with ID: c44fb24f-1bd7-46f9-8581-dd0cde46d490
The registered system name is: ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com
```
So is `--org` a required argument?

And it looks like the module `server_hostname` is *not* at hostname, it is supposed to be a url: https://github.com/ansible-collections/community.general/blob/main/plugins/modules/redhat_subscription.py#L416
```
        if server_hostname:
            args.extend(['--serverurl', server_hostname])
```

richm:
this worked:
```
 - name: Call subscription-manager
   redhat_subscription:
     state: "{{ rhc_state | d('present') }}"
     username: "{{ rhc_auth.login.username | d(omit) }}"
     password: "{{ rhc_auth.login.password | d(omit) }}"
     activationkey: "{{ rhc_auth.activation_keys.keys | join(',')
-  if ( rhc_auth.activation_keys.keys is defined ) else omit }}"
-    org_id: "{{ rhc_auth.activation_keys.org | d(omit) }}"
+      if ( rhc_auth.activation_keys.keys is defined ) else omit }}"
+    org_id: "{{ rhc_server.org | d(omit) }}"
+    server_hostname: "{{ 'https://' ~ rhc_server.hostname ~ __port_val if
+      rhc_server.hostname is defined else omit }}"
+    server_port: "{{ rhc_server.port | d(omit) }}"
+    server_prefix: "{{ rhc_server.prefix | d(omit) }}"
+    server_insecure: "{{ rhc_server.insecure | d(omit) }}"
+  vars:
+    __port_val: "{{ ':' ~ rhc_server.port if rhc_server.port is defined
+      else '' }}"

--- a/tests/tests_register_unregister.yml
+++ b/tests/tests_register_unregister.yml
@@ -3,7 +3,6 @@
 - name: Basic register/unregister
   hosts: all
   gather_facts: true
-  become: true
   tags:
     - tests::slow
   tasks:
@@ -23,6 +22,7 @@
           port: "{{ lsr_rhc_test_data.candlepin_port }}"
           prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
           insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+          org: "{{ lsr_rhc_test_data.reg_organization }}"
 
     - name: Register (noop)
       include_role:
@@ -37,6 +37,7 @@
           port: "{{ lsr_rhc_test_data.candlepin_port }}"
           prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
           insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+          org: "{{ lsr_rhc_test_data.reg_organization }}"
 
     - name: Unregister
       include_role:
```
result
```
PLAY RECAP *********************************************************************
/home/rmeggins/.cache/linux-system-roles/rhel-8-y.qcow2 : ok=51   changed=20   unreachable=0    failed=0    skipped=30   rescued=0    ignored=0   

Wednesday 02 November 2022  16:48:24 -0600 (0:00:00.765)       0:08:36.356 **** 
=============================================================================== 
candlepin : Execute deploy script ------------------------------------- 319.68s
/home/rmeggins/linux-system-roles/rhc/tests/roles/candlepin/tasks/common/cp_deploy.yml:27 
candlepin : Install Candlepin requirements ----------------------------- 92.91s
/home/rmeggins/linux-system-roles/rhc/tests/roles/candlepin/tasks/el8/el8_base_deps.yml:32 
Install system requirements of candlepin role -------------------------- 26.60s
/home/rmeggins/linux-system-roles/rhc/tests/tasks/setup_candlepin.yml:36 ------
```
as expected, the longest tasks are for deploying candlepin and installing the deps/packages.  But 8 minutes 36 seconds is not too bad.

ptoscano:
> > > Yes, it is up to the users, but we want to encourage users to use best security practices. A user new to Ansible might not be aware that Vault exists and secrets should be secured.
> > > next - the various `rhsm` modules require the subscription-manager package:
> > 
> > That opens an interesting question (at least IMHO): if I'm registering a new RHEL system, that means there are no configured/enabled repositories where to get content from, so installing things won't work immediately. In RHEL `subscription-manager` is part of the "core" of a system, so it's kind of essential.
> 
> Good point. I guess a note in the README that either
> 
>     * `subscription-manager` should be installed as part of system provisioning
> 
>     * some repo containing `subscription-manager` should be enabled

I thought about it a bit more, and I think having the:

> ```
> +- name: Ensure required packages are installed
> +  package:
> +    name: "{{ __rhc_packages }}"
> +    state: present
> ```

... as I added it should be enough, with no need to explain it more:
- in case of RHEL, `subscription-manager` is already installed, so it is effectively a no-op
- CentOS/Fedora/etc come with already configured repositories, so installing `subscription-manager` is possible with no issues

> (NOTE: Directives like `become: true` should _not_ be put in the playbooks - instead, users should specify these on the command line as needed e.g. `ansible-playbook -b ...`)

Interesting, I did not know that. Do you have more details/links/docs on this?

> So is `--org` a required argument?

Only in case the user belongs to more than one organization.

... which is the case of the "admin" user in the Candlepin test data; will amend this.

> I made some more changes - now I get to the point where the redhat_subscription module is executed
> 
> ```
>      org_id: "{{ rhc_auth.activation_keys.org | d(omit) }}"
> +    server_hostname: "{{ rhc_server.hostname | d(omit) }}"
> +    server_port: "{{ rhc_server.port | d(omit) }}"
> +    server_prefix: "{{ rhc_server.prefix | d(omit) }}"
> +    server_insecure: "{{ rhc_server.insecure | d(omit) }}"
> 
> 
> However, the `redhat_subscription` module is hung. I logged into the machine and found the subscription-manager command:
> 
> ```
> /usr/libexec/platform-python /usr/sbin/subscription-manager register --insecure --serverurl 127.0.0.1 --username admin --password admin
> ```
> 
> This is wrong.

Yes and no :)

This actually works because the URL parsed from `--serverurl` can have "missing parts" (e.g. no ports and no prefix, like shown above), and the missing bits will be taken from the configuration (`rhsm.conf`).

The `redhat_subscription` module first configures `rhsm.conf` using the supplied `server_*` and `rhsm_*` keys (including the ones mentioned above), see `Rhsm.configure()`... but then `Rhsm.register()` is called with `server_hostname` as it was passed as parameter to `redhat_subscription`, and that is passed to `subscription-manager` as `--serverurl`. This behaviour is definitely confusing, and my explanation is that in the past there used to be only `server_hostname` as "the whole thing", however I need to dig more in the history.

I suspect it's hung for you because of the `--org` thing mentioned earlier. If any of the connection/server parameters was wrong, then `subscription-manager` would fail to connect right away [*].

[*] except possible DNS/port timeouts

richm:
>>    (NOTE: Directives like become: true should not be put in the playbooks - instead, users should specify these on the command line as needed e.g. ansible-playbook -b ...)

> Interesting, I did not know that. Do you have more details/links/docs on this?

Not exactly, but we have users that want to explicitly control the `become` behavior, and it is much easier if the playbooks do not have this.  That is, it is easier for users to specify `-b` on the command line or in the inventory if they want to use become, rather than having to disable it if the playbook has hardcoded `become: true`.

> I suspect it's hung for you because of the --org thing mentioned earlier. 

OK.  It sounds like you know what to do here.

ptoscano:
[citest]

ptoscano:
[citest]

richm:
The candlepin setup will likely fail until I can get https://gitlab.com/testing-farm/gluetool-modules/-/merge_requests/279 merged, then you'll need to add a `tests/collection-requirements.yml` which will install the dependencies needed by candlepin

richm:
> The candlepin setup will likely fail until I can get https://gitlab.com/testing-farm/gluetool-modules/-/merge_requests/279 merged, then you'll need to add a `tests/collection-requirements.yml` which will install the dependencies needed by candlepin

In the meantime, I'd like to merge this - if you can verify that it works locally for you, we can revisit the CI testing later.

ptoscano:
> The candlepin setup will likely fail until I can get https://gitlab.com/testing-farm/gluetool-modules/-/merge_requests/279 merged, then you'll need to add a `tests/collection-requirements.yml` which will install the dependencies needed by candlepin

hm shouldn't `tests/collection-requirements.yml` be installed in addition to `meta/collection-requirements.yml`? this way:
- the "meta" one contains the runtime dependencies of the role
- the "tests" one the additional collections needed by the tests, without duplicating what is specified in the "meta" one

Also, I guess we need a similar change in tox-lsr too? I can do that.

ptoscano:
> > > The candlepin setup will likely fail until I can get https://gitlab.com/testing-farm/gluetool-modules/-/merge_requests/279 merged, then you'll need to add a `tests/collection-requirements.yml` which will install the dependencies needed by candlepin
> > 
> > 
> > hm shouldn't `tests/collection-requirements.yml` be installed in addition to `meta/collection-requirements.yml`?
> 
> Yes. I believe that's what the MR is doing

Oh right, sorry -- reading code at 11pm is generally not a good idea indeed...

> > Also, I guess we need a similar change in tox-lsr too? I can do that.

This is https://github.com/linux-system-roles/tox-lsr/pull/93

richm:
ok - lgtm - please confirm that it works locally with tox-lsr qemu and I'll merge

ptoscano:
> ok - lgtm - please confirm that it works locally with tox-lsr qemu and I'll merge

Yes, it works fine for me in both self-deployed Candlepin and external one, with different distros.

richm (approved):

PR #7, opened 2022-11-09
Improve OS support/checking

ptoscano:
- skip OSes not supported by Candlepin when deploying it
- claim to support EL7 and EL8 for now

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -14,6 +14,12 @@
 
 - name: Deploy Candlepin
   block:
+    - name: Skip on OSes not supported by Candlepin
+      meta: end_play
+      when:
+        - ansible_distribution in ["CentOS", "RedHat"]
+        - ansible_distribution_major_version | int > 8
+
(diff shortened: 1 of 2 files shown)


tests/tasks/setup_candlepin.yml
```diff
@@ -14,6 +14,12 @@
 
 - name: Deploy Candlepin
   block:
+    - name: Skip on OSes not supported by Candlepin
+      meta: end_play
+      when:
+        - ansible_distribution in ["CentOS", "RedHat"]
+        - ansible_distribution_major_version > '8'
```
richm, line 21:
```suggestion
        - ansible_distribution_major_version | int > 8
```
Better to "cast" the version to an `int` rather than rely on string comparison/collation
so candlepin is only supported (currently) on el7 and el8?

ptoscano: >  so candlepin is only supported (currently) on el7 and el8?

not yet, there is work ongoing though; it will be easy to test again when candlepin supports el9

richm (approved):

PR #8, opened 2022-11-11
Perform auto-attach when registering

ptoscano:
This makes entitlement mode working OOTB (at least in the majority of the cases) with no need to fiddle with pools. SCA works regardless of it.

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -18,3 +18,5 @@
     server_port: "{{ rhc_server.port | d(omit) }}"
     server_prefix: "{{ rhc_server.prefix | d(omit) }}"
     server_insecure: "{{ rhc_server.insecure | d(omit) }}"
+    auto_attach: "{{ true
+      if ( rhc_state | d('present') == 'present' ) else omit }}"


tasks/subscription-manager.yml
```diff
@@ -18,3 +18,5 @@
     server_port: "{{ rhc_server.port | d(omit) }}"
     server_prefix: "{{ rhc_server.prefix | d(omit) }}"
     server_insecure: "{{ rhc_server.insecure | d(omit) }}"
+    auto_attach: "{{ true
```
richm, line 21: Do you still want to allow the user to enable or disable auto_attach?  e.g. what if they want to register their system, but want to "manual" (as opposed to auto) attach the system?  If not, then ok.

If so, then perhaps something like this:
```yaml
    auto_attach: "{{ rhc_server.auto_attach if rhc_server.auto_attach is defined
      else true if ( rhc_state | d('present') == 'present' ) else omit }}"
```

ptoscano: > Do you still want to allow the user to enable or disable auto_attach? e.g. what if they want to register their system, but want to "manual" (as opposed to auto) attach the system? If not, then ok.

ATM the goal is to focus on SCA (Simple Content Access), which does not require attachment of pools (neither manual nor automatic). In this case, enabling auto-attach is basically a no-op for SCA, while making things "work" for most of entitlement-based setups.

So, while I don't want to "officially support" entitlement mode for now, at least this "helps" a bit (and also helps testing).

richm (approved):

PR #9, opened 2022-11-11
tests: install some product certificates of deployed Candlepin

ptoscano:
This way it will be possible to use the test products available with the test data of the deployed Candlepin.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -53,6 +53,21 @@
         name: tomcat
         enabled: true
 
+    - name: Ensure /etc/pki/product exists
+      file:
+        path: /etc/pki/product
+        state: directory
+        mode: 0755
+
+    - name: Copy product certificates
+      copy:
+        src: "/home/candlepin/devel/candlepin/generated_certs/{{ item }}.pem"
+        dest: "/etc/pki/product/{{ item }}.pem"
+        remote_src: true
+        mode: 0644
+      loop:
+        - "5050"
+
     - name: Set variables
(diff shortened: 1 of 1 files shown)


tests/tasks/setup_candlepin.yml
```diff
@@ -53,6 +53,20 @@
...
+      file:
+        path: /etc/pki/product
+        state: directory
+        mode: 0755
+
+    - name: Copy product certificates
+      copy:
+        src: "/home/candlepin/devel/candlepin/generated_certs/{{ item }}.pem"
+        dest: "/etc/pki/product/{{ item }}.pem"
+        remote_src: true
```
richm, line 66: you need a `mode` here to fix the ansible-lint error

Does installing candlepin generate some sort of default certs?

ptoscano: > you need a `mode` here to fix the ansible-lint error

ack, added, thanks!

btw: should those lints be run automatically for PRs?

> Does installing candlepin generate some sort of default certs?

The product certificates of the test data are left in the sources, which is where they are copied from (only one for now).

richm: > > you need a `mode` here to fix the ansible-lint error
> 
> ack, added, thanks!
> 
> btw: should those lints be run automatically for PRs?

Yes.  They are run automatically.

> 
> > Does installing candlepin generate some sort of default certs?
> 
> The product certificates of the test data are left in the sources, which is where they are copied from (only one for now).

ok

richm (approved):

PR #11, opened 2022-11-15
add weekly-ci action

ukulekek:
Signed-off-by: Evgeny Fedin <efedin@redhat.com>

### DIFF
--- a/.github/workflows/weekly_ci.yml
+++ b/.github/workflows/weekly_ci.yml
@@ -1,0 +1,79 @@
+# yamllint disable rule:line-length
+name: Weekly CI trigger
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  schedule:
+    - cron: '0 19 * * 6'
+env:
+  BRANCH_NAME: "weekly-ci"
+  COMMIT_MESSAGE: "This PR is to trigger periodic CI testing"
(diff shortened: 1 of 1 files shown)


ukulekek:
This action will trigger the CI system to test the role every weekend. It will ensure the role didn't break because of dependencies and other factors while being inactively developed. See https://url.corp.redhat.com/1ab3ab1 for more details

ptoscano:
can you please rebase this on current `main`? it should work fine now

ukulekek:
Thanks, @ptoscano. I need to rewrite it a bit before merging. See RHELPLAN-132773

ukulekek:
The PR is updated to the latest version @richm

ptoscano:
Thanks! Can you please rebase it on top of `main`? It is slightly behind it.

I see lot of code that is copied among the same actions in other LSR repos; while I don't want to block on this, I think maybe it would be better to create a new github action, so it can be used e.g. as:
```yaml
      - uses: linux-system-roles/periodic-check@v1
        with:
          branch-name: "weekly-ci"  [ maybe optional ]
          commit-message: "This PR ..."  [ maybe optional ]
          body-message: "This PR ..."  [ maybe optional ]
          comment: "[citest]"  [ maybe optional ]
```

ukulekek:
@ptoscano I rebased it. That is a great idea. We can uniform it in future.

PR #12, opened 2022-11-15
Implement rhc_baseurl

ptoscano:
Add a parameter for specifying the base URL for the content received from the subscription server.

This is set as separate parameter on its own, because the redhat_subscription module both configures it and passes it as argument to "subscription-manager register --baseurl"; the problem with `--baseurl` is that subscription-manager parses it and then assumes https as protocol, breaking the original URL set by the user.

Extend the test setup so it is possible to specify it.

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -6,6 +6,19 @@
     state: present
   when: rhc_state | d('present') == "present"
 
+- name: Change rhsm.baseurl
+  command:
+    argv:
+      - subscription-manager
+      - config
+      - "--rhsm.baseurl={{ rhc_baseurl }}"
+  changed_when: false
+  when:
+    - rhc_state | d("present") == "present"
+    - rhc_baseurl | d("") != ""
+    - rhc_baseurl is not none
+    - rhc_baseurl != omit
+
 - name: Call subscription-manager
   redhat_subscription:
     state: "{{ rhc_state | d('present') }}"
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ rhc_server:
 - `insecure` specifies whether to disable the validation of the SSL certificate
   of the registration server
 
+    rhc_baseurl: ""
+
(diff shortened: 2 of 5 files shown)


tasks/subscription-manager.yml
```diff
@@ -6,6 +6,17 @@
     state: present
   when: rhc_state | d('present') == "present"
 
+- name: Change rhsm.baseurl
+  command:
+    argv:
+      - subscription-manager
+      - config
+      - "--rhsm.baseurl={{ rhc_baseurl }}"
```
richm, line 14:
```suggestion
      - "--rhsm.baseurl={{ rhc_baseurl | quote }}"
```

tasks/subscription-manager.yml
```diff
@@ -6,6 +6,17 @@
...
+  changed_when: false
+  when:
+    - rhc_state | d('present') == "present"
```
richm, line 17:
```suggestion
    - rhc_state | d("present") == "present"
```

richm:
@ptoscano ci is failing - will need to fix - but in the meantime, can you include the following change in this PR?  https://github.com/linux-system-roles/rhc/blob/main/.github/workflows/tox.yml#L7
```
-  TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@main"
+   TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@2.13.0"
```

ptoscano:
> @ptoscano ci is failing - will need to fix - but in the meantime, can you include the following change in this PR? https://github.com/linux-system-roles/rhc/blob/main/.github/workflows/tox.yml#L7
> 
> ```
> -  TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@main"
> +   TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@2.13.0"
> ```

Done with #14 (kept the tox-lsr support for `tests/collection-requirements.yml`)

richm (approved):

PR #13, opened 2022-11-15
tests: enable content from deployed Candlepin

ptoscano:
When deploy Candlepin in the test system, tell it to also configure the repositories with test content for the test data (i.e. "-r" for the deploy script). This way it will be possible to simulate fetching actual content from it.

Fetch the GPG key for the repositories, so RPM can install the test packages.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -46,7 +46,7 @@
           vars:
             cp_configure_mariadb: false
             cp_configure_ruby: false
-            cp_deploy_args: "-g -a -f -t"
+            cp_deploy_args: "-g -a -f -t -r"
 
     - name: Enable tomcat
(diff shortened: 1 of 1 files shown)


tests/tasks/setup_candlepin.yml
```diff
@@ -68,6 +68,9 @@
       loop:
         - "5050"
 
+    - name: Install key for RPM repositories
+      shell: curl http://127.0.0.1:8080/RPM-GPG-KEY-candlepin > /etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
```
richm, line 72: can you use the `get_url` module for this instead of `curl`?  https://docs.ansible.com/ansible/2.9/modules/get_url_module.html#get-url-module

richm (approved):

PR #15, opened 2022-11-16
Implement basic registration/unregistration

ahitacat:
Add a minimal task to call "insights-client" via command module to register and unregister.

Extend the API with basic parameters:
- add "rhc_insights_state" for the registration of insights.

Add a very simple integration test that tries to register and unregister. Each step is done twice to verify the idempotency of the role on status changes.

Signed-off-by: Alba Hita Catala <ahitacat@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,14 @@ rhc_repositories:
 
 A release to set for the system. Use `{"state":"absent"}` to actually unset the
 release set for the system.
+```
+rhc_insights:
+  state: present
+```
+
+Whether the system is connected to Insights; valid values are `present`
+(to ensure registration/connection), and `absent`.
+
 
     rhc_proxy: {}
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,6 +2,8 @@
 ---
 rhc_auth: {}
 rhc_baseurl: null
+rhc_insights:
+  state: present
 rhc_organization: null
 rhc_proxy: {}
 rhc_release: null
--- a/tasks/insights-client-unregistration.yml
+++ b/tasks/insights-client-unregistration.yml
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Call insights status
+  include_tasks: check-insights-status.yml
+
+- name: Unregister insights-client
+  command: insights-client --unregister
+  when:
+    - >-
+      "This host is registered" in __rhc_insights_status.stdout
+      or "Registered" in __rhc_insights_status.stdout
+  register: __rhc_insights_unregister_result
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Ensure required packages are installed
+  package:
+    name: insights-client
+    state: present
+
+- name: Call insights status
+  include_tasks: check-insights-status.yml
+
+- name: Register insights-client
+  command: insights-client --register
+  when:
+    - >-
+      "NOT" in __rhc_insights_status.stdout
+      or "unregistered" in __rhc_insights_status.stdout
+  register: __rhc_insights_registration_result
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -3,5 +3,21 @@
 - name: Set ansible_facts required by role
   include_tasks: set_vars.yml
 
+- name: Handle insights unregistration
+  include_tasks: insights-client-unregistration.yml
+  when:
+  - ansible_distribution == "RedHat"
+  - >-
+    rhc_insights.state | d("present") == "absent"
+    or rhc_state | d("present") == "absent"
+  - '"insights-client" in ansible_facts.packages'
+
 - name: Handle system subscription
   include_tasks: subscription-manager.yml
+
+- name: Handle insights registration
+  include_tasks: insights-client.yml
+  when:
+  - ansible_distribution == "RedHat"
+  - rhc_insights.state | d("present") == "present"
+  - rhc_state | d("present") == "present"
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -5,3 +5,11 @@
     gather_subset: min
   when: not ansible_facts.keys() | list |
     intersect(__rhc_required_facts) == __rhc_required_facts
+
+- name: Check if insights-client is installed
+  package_facts:
+  when:
+    - ansible_distribution == "RedHat"
+    - >-
+      rhc_insights.state | d("present") == "absent"
+      or rhc_state | d("present") == "absent"
--- a/tests/tests_insights_client_register.yml
+++ b/tests/tests_insights_client_register.yml
@@ -1,16 +1,24 @@
 # SPDX-License-Identifier: MIT
 ---
-- name: Basic register/unregister
+- name: Basic insights-client register/unregister
   hosts: all
   gather_facts: true
-  become: true
-  tags:
-    - tests::slow
+
   tasks:
-    - name: Setup Candlepin
-      import_tasks: tasks/setup_candlepin.yml
+    - name: Skip when running with embedded Candlepin
+      meta: end_play
+      when:
+        - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
+      delegate_to: 127.0.0.1
+      become: false
 
-    - name: Register
+    - name: Import test data
+      include_vars:
+        file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Register insights
       include_role:
         name: linux-system-roles.rhc
       vars:
@@ -19,7 +27,7 @@
             username: "{{ lsr_rhc_test_data.reg_username }}"
             password: "{{ lsr_rhc_test_data.reg_password }}"
         rhc_insights:
-          state: absent
+          state: present
         rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
         rhc_server:
           hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -27,7 +35,7 @@
           prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
           insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
 
-    - name: Register (noop)
+    - name: Register insights (noop)
       include_role:
         name: linux-system-roles.rhc
       vars:
@@ -36,7 +44,7 @@
             username: "{{ lsr_rhc_test_data.reg_username }}"
             password: "{{ lsr_rhc_test_data.reg_password }}"
         rhc_insights:
-          state: absent
+          state: present
         rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
         rhc_server:
           hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -44,13 +52,13 @@
           prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
           insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
 
-    - name: Unregister
+    - name: Unregister insights
       include_role:
         name: linux-system-roles.rhc
       vars:
         rhc_state: absent
 
-    - name: Unregister (noop)
+    - name: Unregister insights (noop)
       include_role:
         name: linux-system-roles.rhc
       vars:
--- a/tests/tests_release.yml
+++ b/tests/tests_release.yml
@@ -25,6 +25,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -67,6 +69,8 @@
               include_role:
                 name: linux-system-roles.rhc
               vars:
+                rhc_insights:
+                  state: absent
                 # look like a valid release number, hopefully it will never
                 # be used!
                 rhc_release: "1.99"
@@ -84,6 +88,8 @@
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights:
+              state: absent
             rhc_release: "{{ lsr_rhc_test_data.release }}"
 
         - name: Get set release
@@ -98,6 +104,8 @@
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights:
+              state: absent
             rhc_release: "{{ lsr_rhc_test_data.release }}"
 
       always:
--- a/tests/tests_repositories.yml
+++ b/tests/tests_repositories.yml
@@ -20,6 +20,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -57,6 +59,8 @@
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights:
+              state: absent
             rhc_repositories:
               - {name: "*", state: disabled}
 
--- a/tasks/check-insights-status.yml
+++ b/tasks/check-insights-status.yml
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Call insights status
+  command: insights-client --status
+  changed_when: false
+  register: __rhc_insights_status
+  failed_when: __rhc_insights_status.rc not in [0,1]
--- a/tests/tests_proxy.yml
+++ b/tests/tests_proxy.yml
@@ -23,6 +23,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -51,6 +53,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -79,6 +83,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -105,6 +111,8 @@
           login:
             username: "{{ lsr_rhc_test_data.reg_username }}"
             password: "{{ lsr_rhc_test_data.reg_password }}"
+        rhc_insights:
+          state: absent
         rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
         rhc_server:
           hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -136,6 +144,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -164,6 +174,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -194,6 +206,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -224,6 +238,8 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              state: absent
             rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
             rhc_server:
               hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -252,6 +268,8 @@
           login:
             username: "{{ lsr_rhc_test_data.reg_username }}"
             password: "{{ lsr_rhc_test_data.reg_password }}"
+        rhc_insights:
+          state: absent
         rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
         rhc_server:
           hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
@@ -278,6 +296,8 @@
           login:
             username: "{{ lsr_rhc_test_data.reg_username }}"
             password: "{{ lsr_rhc_test_data.reg_password }}"
+        rhc_insights:
+          state: absent
         rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
         rhc_server:
           hostname: "{{ lsr_rhc_test_data.candlepin_host }}"


ptoscano (changes requested):
Good start! I left some initial comments; there are some yamllint issues (mostly empty spaces at the end of lines) that need to be addressed too.

One additional thing is also adapting the existing `tests/tests_register_unregister.yml` to not perform Insights registration (since it is tested by the test added by this PR).

tasks/insights-client.yml
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Ensure required packages are installed
+  package:
+    name: "insights-client"
+    state: present
+  when: rhc_insights_state | d('present') == "present"
+
+- name: Call insights status
+  command: "insights-client --status"
```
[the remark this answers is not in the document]
richm: I suppose, but it is quite common to use the single line command for readability.
You do not have to use double quotes here - basically, in YAML, everything after the `:` is treated as a single string, except when it looks like a number, a boolean, a list, or a dict.
If you need to ensure a variable is quoted propertly, use the `quote` filter:
```
command: insights-client --some-flag={{ rhc_variable | quote }}
```
if you need to wrap long lines, use the YAML string `>-` operator:
```
command: >-
  insights-client --some-flag={{ rhc_variable | quote }}  a very very very very very very very very
  very very very very long line
```

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,28 @@
...
+---
+- name: Basic insights-client register/unregister
+  hosts: all
+  gather_facts: true
+  become: true
+  tags:
+    - tests::slow
+  tasks:
+    - name: Register insights
+      include-role:
```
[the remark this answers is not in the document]
richm: Basically, in ansible, everything on the left hand side of the `:` must be `[a-zA-Z0-9_]` - no `-` allowed. (This isn't strictly true, but it is a good rule of thumb to remember to use `foo_bar:` instead of `foo-bar:`

richm:
please also check the yamllint errors - they are provided in-line in the files in this PR - you can also view them in the `tox / python (3.6) (pull_request)` Details link - https://github.com/linux-system-roles/rhc/actions/runs/3480412613/jobs/5821246153

README.md
```diff
@@ -77,6 +77,12 @@ rhc_server:
 - `insecure` specifies whether to disable the validation of the SSL certificate
   of the registration server
 
+```
+rhc_insights_state: present
```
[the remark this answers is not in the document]
ahitacat: Oh you were right

tasks/insights-client.yml
```diff
@@ -0,0 +1,29 @@
...
+  ignore_errors: true
+  changed_when: false
+  register: __rhc_insights_status
+
+- name: Unregister insights-client
+  command: "insights-client --unregister"
+  when: >- 
+    rhc_insights_state == "absent" 
+    and ("NOT" not in __rhc_insights_status.stdout 
+    or "unregister" not in __rhc_insights_status.stdout)
```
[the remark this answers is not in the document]
ahitacat: I mixed this split with the wrap of long lines

ahitacat:
@ptoscano pushed changes with the modification requested.
I have removed the registration test for insights as they are part of `test_registration_unregistration`

tasks/insights-client.yml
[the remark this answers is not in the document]
ahitacat: Then I keep the command for readability but without the double quotes

README.md
```diff
@@ -77,6 +77,13 @@ rhc_server:
...
```
richm, line 81:
```suggestion
rhc_insights
  state: present
```

defaults/main.yml
```diff
@@ -4,3 +4,5 @@ rhc_auth: {}
 rhc_organization: null
 rhc_server: {}
 rhc_state: "present"
+rhc_insights: 
+  state: "present"
```
richm, line 8:
```suggestion
  state: present
```
The Ansible style is not to use quotes unless you need them.

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Ensure required packages are installed
+  package:
+    name: "insights-client"
```
richm, line 5:
```suggestion
    name: insights-client
```

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
...
+    state: present
+  when: rhc_insights_state | d('present') == "present"
```
richm, line 7:
```suggestion
  when: rhc_insights.state | d("present") == "present"
```

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: Call insights status
+  command: insights-client --status
+  ignore_errors: true
```
richm, line 11: It is far better to use `failed_when` instead of ignoring all errors.  For example - is there some error code which indicates the system is already registered?
If so, then something like this:
```yaml
failed_when: __rhc_insights_status.rc not in [0, 2]
```
assuming `2` means "system already registered".

or, is there some stdout or stderr text which is printed when the system is already registered?  If so, then something like this:
```yaml
failed_when:
  - __rhc_insights_status.rc != 0
  - not __rhc_insights_status.stdout | search("system is already registered")
```

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
...
+  changed_when: false
+  register: __rhc_insights_status
+
+- name: Unregister insights-client
+  command: "insights-client --unregister"
```
richm, line 16:
```suggestion
  command: insights-client --unregister
```

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
...
+  when:
+    - rhc_insights.state == "absent"
```
richm, line 18:
```suggestion
    - rhc_insights.state | d("present") == "absent"
```

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
...
+    - >-
+      "NOT" not in __rhc_insights_status.stdout
+      or "unregister" not in __rhc_insights_status.stdout
```
richm, line 21: See above

tasks/insights-client.yml
```diff
@@ -0,0 +1,31 @@
...
+  register: __rhc_insights_unregister_result
+
+- name: Register insights-client
+  command: insights-client --register
+  when:
+    - rhc_insights.state == "present"
+    - >-
+      "NOT" in __rhc_insights_status.stdout
+      or "unregister" in __rhc_insights_status.stdout
```
richm, line 30: see above

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Basic insights-client register/unregister
+  hosts: all
+  gather_facts: true
+  become: true
```
richm, line 6: you shouldn't put `become: true` here.  If you need to use `become` when you run the playbook locally (you shouldn't if you are using `tox -e qemu-ansible...` as in https://linux-system-roles.github.io/contribute.html, but perhaps you are running against another system), you can use `ansible-playbook -b`

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
...
+  tags:
+    - tests::slow
+  tasks:
+    - name: Unregister insights
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights_state: absent
```
richm, line 14:
```suggestion
        rhc_insights
          state: absent
```

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
...
+
+    - name: Unregister insights (noop)
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights_state: absent
```
richm, line 20:
```suggestion
        rhc_insights
          state: absent
```

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Basic insights-client register/unregister
+  hosts: all
+  gather_facts: true
+  become: true
+  tags:
+    - tests::slow
+  tasks:
+    - name: Unregister insights
```
richm, line 10: Is there a test to register a system with insights?

tasks/insights-client.yml
ahitacat: This case is a bit different. `insights-client` has 2 different returns depending of the `legacy_upload` configuration. That means this can be 0 or 1 and depending of that configuration it means that the command has run without problems or that `insights-client` is register/unregister. 

I want to create a fix to unify this behavior, and once is done remove the `ignore_errors` in the future.

tests/tests_insights_client_register.yml
ahitacat: I removed the test to register insights-client here because it is implicit  in the [test_register_unregister](https://github.com/linux-system-roles/rhc/blob/main/tests/tests_register_unregister.yml)

tests/tests_register_unregister.yml
```diff
@@ -3,9 +3,7 @@
 - name: Basic register/unregister
   hosts: all
   gather_facts: true
-  become: true
-  tags:
-    - tests::slow
```
ptoscano, line 8: this tag flags that the test is slow, which is the case when running it as default without an external candlepin

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Basic insights-client register/unregister
+  hosts: all
+  gather_facts: true
+  become: true
+  tags:
+    - tests::slow
```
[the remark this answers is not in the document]
ahitacat: Yes you are right this one shouldn't be a slow test

tests/tests_insights_client_register.yml
ahitacat: Ok I didn't knew this but I will run locally as you have indicated and remove this line

ahitacat:
I have corrected the conditional for unregistration, it wasn't checking appropriately.

tests/tests_register_unregister.yml
ahitacat: Sorry Pino! I bring back those lines

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
...
+  hosts: all
+  gather_facts: true
+
+  tasks:
+    - name: Unregister insights
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          state: absent
```
ptoscano, line 13: in addition to this, we also need
```yaml
  rhc_state: absent
```
otherwise it will try to register to rhsm (and it fails, because `rhc_auth` was not specified)

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,20 @@
...
+
+    - name: Unregister insights (noop)
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          state: absent
```
ptoscano, line 20: ditto

tests/tests_register_unregister.yml
```diff
@@ -6,6 +6,7 @@
   become: true
   tags:
     - tests::slow
+
```
ptoscano, line 9: extra change :P

tasks/insights-client.yml
```diff
@@ -0,0 +1,30 @@
...
+  ignore_errors: true
+  changed_when: false
+  register: __rhc_insights_status
+
+- name: Unregister insights-client
+  command: insights-client --unregister
+  when:
+    - rhc_insights.state | d("present") == "absent"
+    - ("unregistered" not in __rhc_insights_status.stdout)
+    - ("NOT" not in __rhc_insights_status.stdout)
```
ptoscano, line 20: nitpicks: it is possible to remove the round brackets around the whole conditions

ptoscano (changes requested):
Thanks for the updates! Getting better now.

One small thing first: remember to update the commit message to mention the actual parameter added, as it still mentions `rhc_insights_state`.

Regarding the newly added test `tests_insights_client_register.yml`):
- the name seems slightly confusing, as in reality it only tests the noop unregistration; was it meant to be like this, or it misses something?
- I added a couple of inline notes for fixes that allowed me to run it successfully

We need to extend the test suite to also allow to configure whether to register with Insights or not; in particular, when using the "embedded Candlepin" the Insights support will definitely not work. Not a problem though, as the test suite should easily support this :) check out
- `tests/README.md` that explains the available variables in the `lsr_rhc_test_data` dictionary
- `tests/tasks/setup_candlepin.yml` that sets them for the "embedded Candlepin" case
- the existing `tests/tests_*.yml` that use the `lsr_rhc_test_data.*` variables

In case of "embedded Candlepin", we will need to
- forcefully set `rhc_insights.state: absent`, as there is nothing that can be tested w.r.t. Insights
- skip altogether (using `meta: end_play`) tests that are specific for Insights

I tried to run the existing `tests_register_unregister.yml` test against rhsm.stage:
- the "Register" task ran fine: subscribed against rhsm, `insights-client --status` reported "This host is unregistered.", and then it registered successfully
- during the "Register (noop)" task, the rhsm status did not change (rightfully), but then `insights-client --status` reported again "This host is unregistered.", and thus it registered again
- during the "Unregister" & "Unregister (noop)" tasks, the system was unregistered from rhsm, and then `insights-client --status` reported
    ```
    Unable to fetch egg url https://cert-api.access.redhat.com/r/insights/platform/module-update-router/v1/channel?module=insights-core: 401: Unauthorized. Defaulting to /release
    Please ensure that the system is registered with RHSM for CERT auth, or that correct credentials are set in /etc/insights-client/insights-client.conf for BASIC auth.
    System is registered locally via .registered file. Registered at YYYY-MM-DDTHH:MM:SS.SSSSSS
    Insights API could not be reached to confirm registration status.
    ```
    which caused the "Unregister insights-client" task to not be triggered

The above also lead me to a question: in case of unregistration from Insights, should that be done before even attempting to unregister from (or in general interact with) RHSM?

tasks/insights-client.yml
richm: well, you'll still need quoting - I guess you could try
```yaml
    - "unregistered" not in __rhc_insights_status.stdout
    - "NOT" not in __rhc_insights_status.stdout
```
and if that doesn't work, then use
```yaml
    - '"unregistered" not in __rhc_insights_status.stdout'
    - '"NOT" not in __rhc_insights_status.stdout'
```

richm:
you'll need to rebase on top of the latest `main` branch to pick up the fixes that make tox happy

tests/tests_insights_client_register.yml
ahitacat: `insights-client` needs that rhsm is registered to run. I don't think this is a good option to test the `insights` side.

ptoscano: right :) however this tasks is trying to unregister from insights, and not setting `rhc_state` means it will try to register with rhsm (and fail because of lack of credentials)

ahitacat:
Rebased and modified some things:

1. Added insights registration tests ( to split them from subscription-manager)
2. Split insights logic for unregistration. This needs to be done before subscription-manager gets unregistered.

tasks/insights-client-unregistration.yml
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Check if insights-client is installed
+  ansible.builtin.package_facts:
```
richm, line 4:
```suggestion
  package_facts:
```

tasks/insights-client-unregistration.yml
```diff
@@ -0,0 +1,22 @@
...
+    manager: auto
```
richm, line 5: you can omit this line - `auto` is the default

tasks/insights-client-unregistration.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: Call insights status
+  command: insights-client --status
+  failed_when: false
```
richm, line 9: There is no case when `insights-client --status` will return an error that is unexpected or not able to be handled?
I guess the risk here is that status returns some strange value, and the `Unregister insights-client` task assumes that means the system is not registered, if `unregistered` and `NOT` are not part of the output

For example, if `insights-client --status` returns `0` if registered and `1` if not registered, then you could say 
```
  failed_when: __rhc_insights_status.rc not in [0, 1]
```
or you could use the stdout.  if the output is "System is registered" when registered and "System is unregistered" when unregistered, you could have a failed_when condition like this:
```yaml
  failed_when:
    - '"System is registered" not in __rhc_insights_status.stdout'
    - '"System is unregistered" not in __rhc_insights_status.stdout'
```

tasks/insights-client-unregistration.yml
```diff
@@ -0,0 +1,22 @@
...
+  when:
+    - '"insights-client" in ansible_facts.packages'
+
+- name: Unregister insights-client
+  command: insights-client --unregister
+  when:
+    - '"insights-client" in ansible_facts.packages'
+    - rhc_insights.state | d("present") == "absent"
+    - '"unregistered" not in __rhc_insights_status.stdout'
+    - '"NOT" not in __rhc_insights_status.stdout'
```
richm, line 21: Can you give me examples of the output when the system is registered, and when it is unregistered?

tasks/insights-client.yml
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+---
```
richm, line 2: The code in this file looks very similar to tasks/insights-client-unregistration.yml - can you refactor into a single file?

tasks/insights-client-unregistration.yml
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Check if insights-client is installed
```
richm, line 3: Where do you use the package facts?

ptoscano: the task after this one runs `insights-client --status`, and that is done only if `insights-client` is installed; maybe we need to add the "package" fact to `__rhc_required_facts`?

richm: Unless there is some reason to only conditionally install the insights-clients package, I would just use a `package:` task here:
```yaml
- name: Ensure insights-client is installed
  package:
    name: insights-client
    state: present
```

ptoscano: I guess the idea is to not install install `insights-client` if `rhc_insights.status: absent`. Also, one important thing to note here is that `insights-client` is not installed by default in RHEL (it's in AppStream), while `subscription-manager` is in BaseOS and even part of the `@core` package group (so installed in every RHEL installation); because of this, can be installed `insights-client` only _after_ the system is registered. Hence, I think the tasks before `subscription-manager` runs have to cope with missing packages.

richm: > I guess the idea is to not install install `insights-client` if `rhc_insights.status: absent`.

Ok.

> Also, one important thing to note here is that `insights-client` is not installed by default in RHEL (it's in AppStream), while `subscription-manager` is in BaseOS and even part of the `@core` package group (so installed in every RHEL installation); because of this, `insights-client` can be installed only _after_ the system is registered. Hence, I think the tasks before `subscription-manager` runs have to cope with missing packages.

Why do the tasks in `insights-client-unregistration.yml` have to come before `subscription-manager.yml`?

richm: Is it because if you run `insights-client --unregister` after unsubscribing the system, the command will fail?  If so, is there an error message that can be interpreted to mean "system unsubscribed from rhc" that is different than "system is unregistered with insights" vs. some other error messages?

It seems to me overly complicated to have both `insights-client-unregistration.yml` and `insights-client.yml` with code that is mostly the same.

ptoscano: > Is it because if you run `insights-client --unregister` after unsubscribing the system, the command will fail?

Yes, exactly. Cleanly unregister from Insights before the rhsm unregistration is better than unregistering from rhsm and handling the connection failure of `insights-client` (that cannot authenticate anymore) -- it should properly remove the system from the Insights inventory. The joy of having two tools that need to do closely related tasks and are not aware of each other's presence, you see...

richm:
[citest]

richm:
```
TASK [linux-system-roles.rhc : Call subscription-manager] **********************
task path: /WORKDIR/dist-git-rhc-insights-registration-abevkvup/tests/roles/linux-system-roles.rhc/tasks/subscription-manager.yml:22
Tuesday 06 December 2022  18:15:44 +0000 (0:00:00.084)       0:00:43.901 ****** 
fatal: [sut]: FAILED! => {
    "changed": false
}

MSG:

state is present but any of the following are missing: username, activationkey
	to retry, use: --limit @/tmp/tests_insights_client_register.retry
```
Looks like you need to rebase on top of the latest main to pick up the fix for this.

ptoscano:
> > state is present but any of the following are missing: username, activationkey
> > 	to retry, use: --limit @/tmp/tests_insights_client_register.retry
> 
> Well, no, you already rebased - @ptoscano any idea what is the issue?

Yes, I know about this. The `redhat_subscription` module currently requires either the `username` parameter or the `activationkey` one, in case that `state: present`:
https://github.com/ansible-collections/community.general/blob/b52a7a6b0edcf5f73904259f6a21777cd94a2d49/plugins/modules/redhat_subscription.py#L832

Since we run the module also when there is no status change, the above error message appears in that case. You can see the workaround I added to some tests so far:
```yaml
            # TODO: make it work without rhc_auth
            rhc_auth:
              login:
                username: "{{ lsr_rhc_test_data.reg_username }}"
                password: "{{ lsr_rhc_test_data.reg_password }}"
```
Hence, for now please add the same in the tests where needed. I will soon create a better workaround in the role (so that users don't need to care about this), and also fix it in `redhat_subscription` itself.

tasks/insights-client.yml
ptoscano: Not too much -- I think the biggest duplication here is the query `insights-client --status`. I think we can move that to an helper yml file that registers an internal variable, so the logic is done in only one place. This is similar to what I did in some of the tests, see some of the `tests/tasks/*.yml` files.

tests/tests_insights_client_register.yml
ahitacat: Yes, you are right. I already change it in the last pushed changes

tasks/insights-client.yml
ahitacat: I opted for the second option, the first one didn't work

tasks/insights-client-unregistration.yml
ahitacat: I can limit the error code to be 0 or 1. Also I can change a bit the condition to check if `registered` is present in the output, to not assume it is registered when it returns a 1 and the `unregistered` and `NOT` are not part of the output.

tasks/insights-client-unregistration.yml
ahitacat: It depends of the value of `legacy_upload` that is part of the configuration. The registration output  can be:
```
System is registered locally via .registered file. Registered at 2022-12-05T16:22:19.204489
Insights API confirms registration.
```
or
```
This host is registered
```

And  the unregistration:

```
System is NOT registered locally via .registered file. Unregistered at 2022-12-07T13:29:27.226725
Insights API says this machine is NOT registered.
System unregistered locally via .unregistered file
```
or

```
This host is unregistered
```

ptoscano:
> I will soon create a better workaround in the role (so that users don't need to care about this),

... which is #29

> and also fix it in `redhat_subscription` itself.

... which is https://github.com/ansible-collections/community.general/pull/5664

tasks/insights-client-unregistration.yml
ahitacat: I agree with all @ptoscano has said and also add that the task needs to check if if the package is install to run the commands, otherwise it will fail. 

Additionally, unregistration of `insights-client` needs to be done before unregister `susbscription-manager` it would not fail but is not a clean unregistration. But it needs that `subscription-manager` is registered to register `insights-client`. That is is why it split the logic.

ahitacat:
I added the status as a new tasks, to be called and not duplicating code.

>  One small thing first: remember to update the commit message to mention the actual parameter added, as it still mentions rhc_insights_state.

Done

> the name seems slightly confusing, as in reality it only tests the noop unregistration; was it meant to be like this, or it misses something?

Added the tests that check registration. They were added and removed in the life of this PR.

> tests/README.md that explains the available variables in the lsr_rhc_test_data dictionary
> tests/tasks/setup_candlepin.yml that sets them for the "embedded Candlepin" case
> the existing tests/tests_*.yml that use the lsr_rhc_test_data.* variables
> In case of "embedded Candlepin", we will need to

Done, added 'absent' when doing the unregistration. Thanks to this [PR](https://github.com/linux-system-roles/rhc/pull/29)

tasks/insights-client-unregistration.yml
```diff
@@ -0,0 +1,14 @@
...
+  when: rhc_insights.state | d("present") == "absent"
+
+- name: Unregister insights-client
+  command: insights-client --unregister
+  when:
+    - '"insights-client" in ansible_facts.packages'
+    - rhc_insights.state | d("present") == "absent"
+    - '"This host is registered" in __rhc_insights_status.stdout'
+    - '"Registered" in __rhc_insights_status.stdout'
+  register: __rhc_insights_unregister_result
```
ptoscano, line 14: I think we can use a `block` for both, factorizing some of the conditions:
```yaml
- when:
    - rhc_insights.state | d("present") == "absent"
  block:
    ...
```
Also, adding the `'"insights-client" in ansible_facts.packages'` condition in that block (used by the "Unregister insights-client" task) allows you to remove it from `check-insights-status.yml` (when used again later in `insights-client.yml`, it is executed after `insights-client` is installed, so it will be definitely present).

tasks/insights-client.yml
```diff
@@ -0,0 +1,21 @@
...
+- name: Ensure required packages are installed
+  package:
+    name: insights-client
+    state: present
+  when: rhc_insights.state | d("present") == "present"
```
ptoscano, line 7: I think that, since all the tasks here need to be run when `rhc_insights.state=present`, maybe we can apply the condition to the `include_tasks: insights-client.yml` in `main.yml`; since this helper file is used to register to insights, we can assume this in all the tasks, and simplify things a bit

defaults/main.yml
```diff
@@ -7,3 +7,5 @@ rhc_release: null
 rhc_repositories: []
 rhc_server: {}
 rhc_state: "present"
+rhc_insights:
+  state: present
```
ptoscano, line 11: nitpick: can you please place `rhc_insights` in alphabetical order? :)

ptoscano (changes requested):
Thanks for the updates! And sorry for the extra time, luckily other issues that were on the way also of this work are fixed by now :crossed_fingers:

In addition to the small notes I left, there are couple of things that needs to be done:
- add `rhc_insights.state: absent` to _every_ `linux-system-roles.rhc` usage in all the existing tests: this is because those tests don't test Insights
- skip the newly added `tests_insights_client_register.yml` when running using the embedded Candlepin: there is no Insights server in that case :) you can do it easily with e.g.:
   ```yaml
   - name: Skip when running with embedded Candlepin
     meta: end_play
     when:
       - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
   ```
  as first task in the play

tasks/insights-client-unregistration.yml
ahitacat: Good one! I'm adding blocks to the main tasks

tests/tests_register_unregister.yml
```diff
@@ -18,6 +18,7 @@
           login:
             username: "{{ lsr_rhc_test_data.reg_username }}"
             password: "{{ lsr_rhc_test_data.reg_password }}"
+        rhc_insights.state: absent
```
richm, line 21:
```suggestion
        rhc_insights:
          state: absent
```

tests/tests_register_unregister.yml
```diff
@@ -33,6 +34,7 @@
...
```
richm, line 37:
```suggestion
        rhc_insights:
          state: absent
```

tests/tests_register_unregister.yml
```diff
@@ -44,10 +46,12 @@
       include_role:
         name: linux-system-roles.rhc
       vars:
+        rhc_insights.state: absent
```
richm, line 49:
```suggestion
        rhc_insights:
          state: absent
```

tests/tests_register_unregister.yml
```diff
@@ -44,10 +46,12 @@
...
         rhc_state: absent
 
     - name: Unregister (noop)
       include_role:
         name: linux-system-roles.rhc
       vars:
+        rhc_insights.state: absent
```
richm, line 56:
```suggestion
        rhc_insights:
          state: absent
```

tests/tests_release.yml
```diff
@@ -25,6 +25,7 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights.state: absent
```
richm, line 28:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_release.yml
```diff
@@ -45,6 +46,7 @@
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights.state: absent
```
richm, line 49:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_release.yml
```diff
@@ -59,6 +61,7 @@
...
```
richm, line 64:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_release.yml
```diff
@@ -69,6 +72,7 @@
               vars:
                 # look like a valid release number, hopefully it will never
                 # be used!
+                rhc_insights.state: absent
```
richm, line 75:
```suggestion
                rhc_insights:
                  state: absent
```

tests/tests_release.yml
```diff
@@ -98,11 +103,13 @@
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights.state: absent
```
richm, line 106:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_release.yml
```diff
@@ -98,11 +103,13 @@
...
             rhc_release: "{{ lsr_rhc_test_data.release }}"
 
       always:
         - name: Unregister
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights.state: absent
```
richm, line 114:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_repositories.yml
```diff
@@ -20,6 +20,7 @@
               login:
                 username: "{{ lsr_rhc_test_data.reg_username }}"
                 password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights.state: absent
```
richm, line 23:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_repositories.yml
```diff
@@ -57,6 +58,7 @@
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights.state: absent
```
richm, line 61:
```suggestion
            rhc_insights:
              state: absent
```

tests/tests_repositories.yml
```diff
@@ -76,4 +78,5 @@
...
```
richm, line 81:
```suggestion
            rhc_insights:
              state: absent
```

README.md
```diff
@@ -97,6 +97,14 @@ rhc_repositories:
 
 A release to set for the system. Use `{"state":"absent"}` to actually unset the
 release set for the system.
+```
+rhc_insights:
+  state: present
+```
+
+Whether the system is connected to Insights; valid values are `present`
+(to ensure registration/connection), and `absent`.
```
richm, line 106: What if the user does not specify `rhc_insights` at all?  What should the role do?  All of the `rhc_insights: state: absent` added below suggests that the role will **always** attempt to change the insights registration status.  Is this really what we want?  I mean, if the default is `present`, then the role will always attempt to register with insights, *even if the user does not want that*, even if `rhc_insights` is not specified.  Conversely, if the default is `absent`, then the role will always attempt to unregister with insights if `rhc_insights` is not specified.

I think the role should do like @ptoscano did elsewhere:
* if `rhc_insights` is not set i.e. `rhc_insights == {}`, do nothing - do not attempt to register or unregister
* if `rhc_insights` has `state: present` - register
* if `rhc_insights` has `state: absent` - unregister

defaults/main.yml
```diff
@@ -2,6 +2,8 @@
 ---
 rhc_auth: {}
 rhc_baseurl: null
+rhc_insights:
```
richm, line 5:
```suggestion
rhc_insights: {}
```

defaults/main.yml
```diff
@@ -2,6 +2,8 @@
...
+  state: present
```
richm, line 6:
```suggestion

```

tasks/main.yml
```diff
@@ -3,5 +3,13 @@
...
+- name: Handle insights unregistration
+  include_tasks: insights-client-unregistration.yml
+  when: rhc_insights.state | d("present") == "absent"
+
 - name: Handle system subscription
   include_tasks: subscription-manager.yml
+
+- name: Handle insights registration
+  include_tasks: insights-client.yml
+  when: rhc_insights.state | d("present") == "present"
```
richm, line 15:
```suggestion
  when: rhc_insights.state | d(none) == "present"
```

tasks/set_vars.yml
```diff
@@ -4,3 +4,7 @@
     gather_subset: min
   when: not ansible_facts.keys() | list |
     intersect(__rhc_required_facts) == __rhc_required_facts
+
+- name: Check if insights-client is installed
+  package_facts:
+  when: rhc_insights.state | d("present") == "absent"
```
richm, line 10:
```suggestion
  when: rhc_insights.state | d(none) == "absent"
```

tests/tests_register_unregister.yml
richm, line 21:
```suggestion

```

tests/tests_register_unregister.yml
richm, line 37:
```suggestion

```

tests/tests_register_unregister.yml
richm, line 56:
```suggestion

```

tests/tests_release.yml
richm, line 28:
```suggestion

```

tests/tests_release.yml
richm, line 49:
```suggestion

```

tests/tests_release.yml
richm, line 64:
```suggestion

```

tests/tests_release.yml
```diff
@@ -84,6 +88,7 @@
...
```
richm, line 91:
```suggestion

```

tests/tests_release.yml
richm, line 106:
```suggestion

```

tests/tests_release.yml
richm, line 114:
```suggestion

```

tests/tests_repositories.yml
richm, line 23:
```suggestion

```

tests/tests_repositories.yml
richm, line 81:
```suggestion

```

README.md
ahitacat: I think this is a better option. I was strictly using the definition we made in the document, I'll update that too.

richm: yes, sounds good

ahitacat: Just a note to this. The tests that didn't use insights-client won't try to unregister insights as it is not registered.

>rhc_state=present (specified or default): rhc_insights.state can be anything (defaults to present, so it will either 
> register to Insights or not, still with a registered-to-rhsm system
> rhc_state=absent (specified):
> rhc_insights.state ought to default to absent
> rhc_insights.state=present (specified) ought to be rejected
> rhc_insights.state=absent (specified) is OK

I think this is correct.

tasks/main.yml
```diff
@@ -3,5 +3,13 @@
 - name: Set ansible_facts required by role
   include_tasks: set_vars.yml
 
+- name: Handle insights unregistration
+  include_tasks: insights-client-unregistration.yml
+  when: rhc_insights.state | d('present') == "absent"
```
ptoscano, line 8: - I think we need to limit this also to RHEL, ie `ansible_distribution == "RedHat"`, as `insights-client` exists only there
- in addition to check for `rhc_insights.state`, I think we need to unregister from Insights also when `rhc_state | d("present") == "absent` -- it does not make sense to not unregister from Insights when unregistering from rhsm
   - this will also allow us to drop the manually specified `rhc_insights.state=absent` when `rhc_state=absent` in the tests
- minor nitpick: double quotes instead of single ones

tasks/main.yml
```diff
@@ -3,5 +3,13 @@
...
+
 - name: Handle system subscription
   include_tasks: subscription-manager.yml
+
+- name: Handle insights registration
+  include_tasks: insights-client.yml
+  when: rhc_insights.state | d('present') == "absent"
```
ptoscano, line 15: most likely, the comparison here ought to be with "present" rather than "absent" ;)

also, as I mentioned for the above `Handle insights unregistration task:
- perform this task only on RHEL
- minor nitpick: double quotes instead of single ones

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,41 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Skip when running with embedded Candlepin
+  meta: end_play
+  when:
+    - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
```
ptoscano, line 6: this needs to be moved as first task in `tasks`, as it does not work here:
```
ERROR! 'meta' is not a valid attribute for a Play

The error appears to be in '$PWD/rhc/tests/tests_insights_client_register.yml': line 3, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

---
- name: Skip when running with embedded Candlepin
  ^ here
```
also, since the envvar is in the system where ansible runs (and not on the managed node), it needs also:
```yaml
      delegate_to: 127.0.0.1
      become: false
```

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: Basic insights-client register/unregister
+  hosts: all
+  gather_facts: true
+
+  tasks:
+    - name: Register insights
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
```
ptoscano, line 16: to register properly, this also needs `rhc_auth`, `rhc_organization`, and `rhc_server` -- pick them as available in `tests_register_unregister.yml`; to make them work, you need to import the file from `LSR_RHC_TEST_DATA` with the config -- add this task before the first register task:
```yaml
    - name: Import test data
      include_vars:
        file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
      delegate_to: 127.0.0.1
      become: false
```

ptoscano (changes requested):

tests/tests_insights_client_register.yml
```diff
@@ -0,0 +1,70 @@
...
+      meta: end_play
+      when:
+        - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Import test data
+      include_vars:
+        file: "{{ lsr_rhc_test_data_file }}"
+      when: lsr_rhc_test_data_file | length > 0
```
ptoscano, line 18: `lsr_rhc_test_data_file` does not exist here (`setup_candlepin.yml` saves it as fact, but here that yml is not used) -- simply use the `lookup()` here (/me notes to self: rearrange this stuff a bit)

also, the `when` here is redundant, as always true (the opposite of it in the task before this causes the end of the play execution)

ptoscano (changes requested):
Thanks for the updates! Almost there!

tasks/set_vars.yml
```diff
@@ -5,3 +5,7 @@
     gather_subset: min
   when: not ansible_facts.keys() | list |
     intersect(__rhc_required_facts) == __rhc_required_facts
+
+- name: Check if insights-client is installed
+  package_facts:
+  when: rhc_insights.state | d('present') == "absent"
```
richm, line 11:
```suggestion
  when: rhc_insights.state | d("present") == "absent"
```
please be consistent with quoting - Ansible style prefers double quotes unless the context requires single

tests/tests_release.yml
```diff
@@ -98,11 +108,15 @@
...
+              state: absent
             rhc_release: "{{ lsr_rhc_test_data.release }}"
 
       always:
         - name: Unregister
           include_role:
             name: linux-system-roles.rhc
           vars:
+            rhc_insights:
+              state: absent
```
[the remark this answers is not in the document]
ahitacat: We can omit the insights variable here as well, right?
```            
rhc_insights:
   state: absent
rhc_release: "{{ lsr_rhc_test_data.release }}"
```

tasks/set_vars.yml
```diff
@@ -5,3 +5,7 @@
     gather_subset: min
   when: not ansible_facts.keys() | list |
     intersect(__rhc_required_facts) == __rhc_required_facts
+
+- name: Check if insights-client is installed
+  package_facts:
+  when: rhc_insights.state | d("present") == "absent"
```
ptoscano, line 11: this needs to match the condition of the `Handle insights unregistration` task:
```suggestion
  when:
    - ansible_distribution == "RedHat"
    - >-
      rhc_insights.state | d("present") == "absent"
      or rhc_state | d("present") == "absent"
```

ptoscano (changes requested):
woot! with a simple condition change, now the new test works fine for me :)

tasks/set_vars.yml
ahitacat: Oh right! The tests have change and didn't change the `rhc_insights.state` but the `rhc_state`.

ptoscano:
[citest]

ptoscano:
[citest]

richm:
@ahitacat the proxy tests are failing too - you'll need to add
```yaml
        rhc_insights:
          state: absent
```
in a similar way to the proxy tests that you did to the other tests

ptoscano:
also, on rhel7 it seems that `insights-client --status` exits with rc=2 -- see this output from `tests_register_unregister.yml`:
```
TASK [linux-system-roles.rhc : Call insights status] ***************************
task path: /WORKDIR/dist-git-rhc-insights-registration-_b_aneex/tests/roles/linux-system-roles.rhc/tasks/check-insights-status.yml:3
Tuesday 13 December 2022  19:43:53 +0000 (0:00:00.126)       0:11:52.682 ****** 
fatal: [sut]: FAILED! => {
    "changed": false,
    "cmd": "insights-client --status",
    "failed_when_result": true,
    "rc": 2
}

MSG:

[Errno 2] No such file or directory
	to retry, use: --limit @/tmp/tests_register_unregister.retry
```
adding "2" to the list of allowed `rc` in `tasks/check-insights-status.yml` seems to make it work fine (not sure whether there are side effects though)

ptoscano:
[citest]

ptoscano:
[citest bad]

ptoscano (approved):
Thanks for the work and the patience on this! It looks good to me, so I'm approving this -- will let @richm do another review.

ptoscano:
[citest bad]

ptoscano:
[citest bad]

richm (approved):

PR #17, opened 2022-11-16
Implement rhc_repositories

ptoscano:
Add a parameter for specifying the repositories to enable or disable, and use it to invoke the "rhsm_repository" module accordingly. Note that, even if "rhsm_repository" accepts more than one repository at a time, they all must have the same state, which is not something we can assume from the list of repositories passed as parameter. Also, use "rhsm_repository" in a block for subscribed-only system, as there will be more tasks to be run there in the future.

The only gotcha here is that we need to pass "rhc_auth" also when invoking the "rhc" role only to disable all the repositories: this is because we unconditionally invoke the "redhat_subscription" module in the role, and that module currently require either the "username" or the "activationkey" parameter in case that "state=present" (which is the default). This will need future changes in the role to workaround this limitation, and possibly making the "redhat_subscription" module slightly smarter.

Make sure to test this:
- extend the test setup with test repositories
- add an helper script for querying the list of repositories in a single way, that works like a function
- add a new test to verify enabling & disabling repositories works as expected


tasks/subscription-manager.yml
```diff
@@ -33,3 +33,16 @@
...
+  block:
+    - name: Configure repositories
+      rhsm_repository:
+        name: "{{ item.name }}"
+        state: "{{ item.state }}"
+        purge: no
+      loop: "{{ rhc_repositories }}"
+      when:
+        - ( rhc_repositories | d([]) ) | length > 0
+  when:
```
richm, line 47: The new ansible recommendation is to have the `block/rescue/always` be the last keywords in the task - any `when` or `vars` should come after `name` and before `block`  - ansible-lint 6.x will flag this

tests/list_repositories.yml
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: MIT
+---
+- name: subscription-manager repos --{{ mode | d("list") }}
+  shell: |
```
richm, line 4:
```suggestion
  shell: |
    set -euo pipefail
```
should be flagged by ansible-lint

richm (approved):

PR #19, opened 2022-11-16
Fix checks when setting rhsm.baseurl

ptoscano:
Since the default is null, then check for None or omit values before
checking whether it is undefined or an empty string.

Fix also the empty string check:
- remove the "default()" filter, as it should be defined as role
  default variable
- cast it as string, so numbers will not cause the "length" filter to
  bomb out

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -15,9 +15,9 @@
   changed_when: false
   when:
     - rhc_state | d("present") == "present"
-    - rhc_baseurl | d("") | length > 0
     - rhc_baseurl is not none
     - rhc_baseurl != omit
+    - rhc_baseurl | string | length > 0
 
 - name: Call subscription-manager
   redhat_subscription:


tasks/subscription-manager.yml
```diff
@@ -15,9 +15,9 @@
   changed_when: false
   when:
     - rhc_state | d("present") == "present"
-    - rhc_baseurl | d("") | length > 0
     - rhc_baseurl is not none
     - rhc_baseurl != omit
+    - rhc_baseurl | d("") | length > 0
```
richm, line 20: hmm - this is a little tricky - since `rhc_baseurl` is specified in defaults/main.yml it will **always** be defined - so the `d(..)` filter is not needed - https://github.com/linux-system-roles/rhc/blob/main/defaults/main.yml#L4

In addition - you cannot use the `length` filter on a `none` value:
```
ansible -c local -i localhost, all -m debug -a "echo {{ none | length > 0 }}"
localhost | FAILED! => {
    "msg": "Unexpected templating type error occurred on (echo {{ none | length > 0 }}): object of type 'NoneType' has no len()"
}
```
try this
```yaml
    - rhc_baseurl is not none
    - rhc_baseurl != omit
    - rhc_baseurl | length > 0
```
(yes - some user might do something weird like `rhc_baseurl: 12345` and the `length` would fail with an error, but I don't think we need to worry about that case)

ptoscano: yeah, the juggling between undefined, omit, null & co gave me a little headache earlier today :unamused:

removing `d(..)` -> makes sense to me, initially i thought the same because of the role defaults, then not sure what confused me towards other directions...

regarding the "`rhc_baseurl: 12345` vs `length`" case that you mention: what if i cast the variable to string, eg `rhc_baseurl | string | length > 0`? shouldn't this make it even more safe?

richm: > yeah, the juggling between undefined, omit, null & co gave me a little headache earlier today unamused
> 
> removing `d(..)` -> makes sense to me, initially i thought the same because of the role defaults, then not sure what confused me towards other directions...
> 
> regarding the "`rhc_baseurl: 12345` vs `length`" case that you mention: what if i cast the variable to string, eg `rhc_baseurl | string | length > 0`? shouldn't this make it even more safe?

Sure - go ahead and add a `| string`

richm (approved):

PR #21, opened 2022-11-24
tests: add a failing tests_default

ptoscano:
By design, the role ensures that the system is subscribed by default; to subscribe an unsubscribed system credentials are needed (the "rhc_auth" parameters to the role), so if they are not passed the role will fail.

Add a basic test (without Candlepin) that checks for this situation.


richm:
looks good - please rebase on top of the latest main branch to pick up the changes to make tox happy

richm (approved):

PR #20, opened 2022-11-22
meta: temporarily add tests requirements for the role

ptoscano:
The test infra does not support tests/collection-requirements.yml yet; hence, for now pull the extra collections needed for the tests as dependencies for the role.

PR #23, opened 2022-12-01
Implement rhc_release

ptoscano:
Add a parameter for specifying the release to set for the system: pass it as parameter for "redhat_subscription", which will set the release, however only when registering (i.e. with a status change). When already registered, manually invoke the "rhsm_release" module to set/unset release, making sure to do it only when there is no status change (i.e. system registered, and requesting it to be registered).

The only gotcha here is that we need to pass "rhc_auth" also when invoking the "rhc" role only to disable all the repositories: this is because we unconditionally invoke the "redhat_subscription" module in the role, and that module currently require either the "username" or the "activationkey" parameter in case that "state=present" (which is the default). This will need future changes in the role to workaround this limitation, and possibly making the "redhat_subscription" module slightly smarter.

Make sure to test this:
- extend the test setup with a release variable
- add an helper script for querying in a single way the currently set release, that works like a function
- add a new test to verify that:
  - setting the release works when registering
  - setting/unsetting the release works on an already registered system
  - invalid releases are properly rejected (by subscription-manager)
- skip the test in case it is not possible to set a release


tests/tests_release.yml
```diff
@@ -0,0 +1,107 @@
...
+    - tests::slow
+  tasks:
+    - name: Setup Candlepin
+      import_tasks: tasks/setup_candlepin.yml
+
+    - name: Skip if no test release is set
+      meta: end_play
+      when:
+        - lsr_rhc_test_data.release is not defined
+        - lsr_rhc_test_data.release is none
```
richm, line 17:
```yaml
        - lsr_rhc_test_data.release | d(none) == none
````

**Comment by ptoscano** [comment]
OK, I think I solved the set/unset/unchanged problem, thanks Rich! Also improved a bit the test.

File: tests/tests_release.yml [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+          vars:
+            # TODO: make it work without rhc_auth
+            rhc_auth:
+              login:
+                username: "{{ lsr_rhc_test_data.reg_username }}"
+                password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_release: {"state":"absent"}
+
+        - name: Check invalid releases cannot be set
+          ignore_errors: true
```
richm, line 75: do you need `ignore_errors` here?  The purpose of this test is to generate an error and catch it and see if it is the right error

richm:
[citest]

looks pretty good, one question about ignore_errors

tests/tests_release.yml
ptoscano: Ah, now I know why I had to do that -- my fault: I used "no-such-release" as wrong release, which however was rejected by the `rhsm_release` module rather than `subscription-manager`, and thus the assertion `'No releases match ' in ansible_failed_result.msg` failed. Thanks for the poke!

Fixed the invalid release (with a comment), removed `ignore_errors`, added safeguard `fail` task.

ptoscano:
> looks pretty good, one question about ignore_errors

hopefully fixed

> [citest]

we need #20 for now to fix testing in testing-farm

richm:
ok - rebase on top of the latest main branch to pick up the meta change, then mark as Ready for review

ptoscano:
[citest]

richm (approved):

PR #28, opened 2022-12-07
Implement rhc_proxy

ptoscano:
Add a parameter for configuring the proxy server used for connecting.

The syntax of the 'rhc_proxy' parameter is particular:
- 'rhc_proxy: {"state":"absent"}': resets the proxy configuration to
  the default one (i.e. no proxy)
- if an attribute is not specified in 'rhc_proxy', then it is not
  changed
- if an attribute is specified in 'rhc_proxy', then it is set to that
Because of this, there is some jinja gymnastic needed to all the
'server_proxy_*' parameters of the 'redhat_subscription' module to
either omit, set as empty (using a helper empty variable), or use the
passed value.

Make sure to test this:
- extend the test setup with variables related to unauthenticated,
  authenticated, and non-working proxy servers
- add an helper script to deploy/configure squid in the test system;
  having the proxy server in the same system does not allow the test
  to properly check as if it was on a different system, however this
  is the best that can be done with a all-in-one-system setup
- add a new test to verify:
  - various combinations of non-working proxy details, for both
    unauthenticated, and authenticated proxy servers, lead to
    registration failures
  - using correct details works fine
  - resetting the proxy server works fine too

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,26 @@ rhc_repositories:
 A release to set for the system. Use `{"state":"absent"}` to actually unset the
 release set for the system.
 
+    rhc_proxy: {}
+
+The details of the proxy server to use for connecting:
+```yaml
+rhc_proxy:
+  hostname: "proxy-hostname"
+  port: 4321
+  username: "proxy-hostname"
+  password: "proxy-password"
+```
+- `hostname` is the hostname of the proxy server
+- `port` is the port to which connect to on the proxy server
+- `username` is the username to use for authenticating on the proxy server;
+  it can be not specified if the proxy server does not require authentication
+- `password` is the password to use for authenticating on the proxy server;
+  it can be not specified if the proxy server does not require authentication
+
+Use `{"state":"absent"}` to reset all the proxy configurations to empty
+(effectively disabling the proxy server).
+
 ## Dependencies
 
 None.
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,7 @@
 rhc_auth: {}
 rhc_baseurl: null
 rhc_organization: null
+rhc_proxy: {}
 rhc_release: null
 rhc_repositories: []
 rhc_server: {}
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -54,6 +54,34 @@
            and __rhc_subman_identity.rc == 1
            and rhc_release != __rhc_state_absent )
       else omit }}"
+    server_proxy_hostname: "{% if rhc_proxy == __rhc_state_absent -%}
+      {{ __rhc_empty_string }}
+      {%-   elif 'hostname' not in rhc_proxy -%}
+      {{ omit }}
+      {%-   else -%}
+      {{ rhc_proxy.hostname }}
+      {%- endif %}"
+    server_proxy_port: "{% if rhc_proxy == __rhc_state_absent -%}
+      {{ __rhc_empty_string }}
+      {%-   elif 'port' not in rhc_proxy -%}
+      {{ omit }}
+      {%-   else -%}
+      {{ rhc_proxy.port }}
+      {%- endif %}"
+    server_proxy_user: "{% if rhc_proxy == __rhc_state_absent -%}
+      {{ __rhc_empty_string }}
+      {%-   elif 'username' not in rhc_proxy -%}
+      {{ omit }}
+      {%-   else -%}
+      {{ rhc_proxy.username }}
+      {%- endif %}"
+    server_proxy_password: "{% if rhc_proxy == __rhc_state_absent -%}
+      {{ __rhc_empty_string }}
+      {%-   elif 'password' not in rhc_proxy -%}
+      {{ omit }}
+      {%-   else -%}
+      {{ rhc_proxy.password }}
+      {%- endif %}"
 
 - name: Configure subscribed system
   when:
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -99,6 +99,16 @@
             - {name: "admin-content-label-5051", state: enabled}
             - {name: "content-label-32060", state: disabled}
           release: null  # no releases in Candlepin test data
+          proxy_noauth_hostname: localhost
+          proxy_noauth_port: 3128
+          proxy_auth_hostname: localhost
+          proxy_auth_port: 3130
+          proxy_auth_username: "proxyuser"
+          proxy_auth_password: "proxyuser"
+          proxy_nonworking_hostname: "wrongproxy"
+          proxy_nonworking_port: 4000
+          proxy_nonworking_username: "wrong-proxyuser"
+          proxy_nonworking_password: "wrong-proxypassword"
         cacheable: true
   when: lsr_rhc_test_data_file | length == 0
 
--- a/tests/tasks/setup_squid.yml
+++ b/tests/tasks/setup_squid.yml
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Install and configure squid
+  when:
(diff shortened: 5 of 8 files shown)


ptoscano:
[citest]

README.md
```diff
@@ -98,6 +98,26 @@ rhc_repositories:
 A release to set for the system. Use `{"state":"absent"}` to actually unset the
 release set for the system.
 
+    rhc_proxy: {}
+
+The details of the proxy server to use for connecting:
+```yaml
+rhc_server:
```
richm, line 105:
```suggestion
rhc_proxy:
```

tasks/subscription-manager.yml
```diff
@@ -47,6 +47,38 @@
            and __rhc_subman_identity.rc == 1
            and rhc_release != __rhc_state_absent )
       else omit }}"
+    server_proxy_hostname: "{% if rhc_proxy is none -%}
+      {{ omit }}
+      {%- elif rhc_proxy == {} -%}
+      {%- elif 'hostname' not in rhc_proxy -%}
```
richm, line 53:
```suggestion
      {%-   elif 'hostname' not in rhc_proxy -%}
```
indent for readability

tasks/subscription-manager.yml
```diff
@@ -47,6 +47,38 @@
            and __rhc_subman_identity.rc == 1
            and rhc_release != __rhc_state_absent )
       else omit }}"
+    server_proxy_hostname: "{% if rhc_proxy is none -%}
+      {{ omit }}
+      {%- elif rhc_proxy == {} -%}
```
richm, line 52: you have an empty `elif` clause here?

tests/tasks/setup_candlepin.yml
```diff
@@ -99,6 +99,16 @@
             - {name: "admin-content-label-5051", state: enabled}
             - {name: "content-label-32060", state: disabled}
           release: null  # no releases in Candlepin test data
+          proxy_noauth_hostname: 127.0.0.1
```
richm, line 102: do you need to use `localhost` here instead of `127.0.0.1`?

tests/tasks/setup_squid.yml
```diff
@@ -0,0 +1,88 @@
...
+      lineinfile:
+        path: /etc/squid/squid.conf
+        regexp: "^acl SSL_ports port {{ lsr_rhc_test_data.candlepin_port }}"
+        insertbefore: "^acl Safe_ports "
+        line: >-
+          acl SSL_ports port {{ lsr_rhc_test_data.candlepin_port }} # Candlepin
+
+    - name: Setup no authentication
+      when:
+        - authentication | d(false) == false
```
richm, line 47:
```suggestion
        - not (authentication | d(false))
```
ansible/jinja style is not to do comparisons to `true` or `false`

tasks/subscription-manager.yml
ptoscano: it's on purpose: if the dictionary is empty, it means "reset every proxy attribute to their default", and for all of the ones added the default is "empty string"

richm: No, I meant that you have an `elif` and there is nothing done - I've never seen this construct in jinja before - I guess it is an implicit no-op?  something like
```
elif rhc_proxy == {}:
  pass
elif ....
```

richm:
[citest]

richm:
The test
```
TASK [Register (no authentication)] ********************************************
task path: /home/rmeggins/linux-system-roles/rhc/tests/tests_proxy.yml:100
Thursday 08 December 2022  14:11:03 -0700 (0:00:00.018)       0:09:46.546 ***** 
```
fails here
```
TASK [linux-system-roles.rhc : Call subscription-manager] **********************
task path: /home/rmeggins/linux-system-roles/rhc/tests/roles/linux-system-roles.rhc/tasks/subscription-manager.yml:31
Thursday 08 December 2022  14:11:06 -0700 (0:00:00.024)       0:09:48.908 ***** 
fatal: [/home/rmeggins/.cache/linux-system-roles/rhel-8-y.qcow2]: FAILED! => {
    "changed": false,
    "cmd": "/usr/sbin/subscription-manager register --org VALUE_SPECIFIED_IN_NO_LOG_PARAMETER --auto-attach --username VALUE_SPECIFIED_IN_NO_LOG_PARAMETER --password '********'",
    "rc": 70
}

STDOUT:

Registering to: localhost:8443/candlepin

STDERR:

Proxy error, unable to connect to proxy server.
```

tasks/subscription-manager.yml
ptoscano: The result is an empty string, which is what I want.

ptoscano:
> The test
> 
> ```
> TASK [Register (no authentication)] ********************************************
> task path: /home/rmeggins/linux-system-roles/rhc/tests/tests_proxy.yml:100
> Thursday 08 December 2022  14:11:03 -0700 (0:00:00.018)       0:09:46.546 ***** 
> ```
> 
> fails here
> 
> ```
> TASK [linux-system-roles.rhc : Call subscription-manager] **********************
> task path: /home/rmeggins/linux-system-roles/rhc/tests/roles/linux-system-roles.rhc/tasks/subscription-manager.yml:31
> Thursday 08 December 2022  14:11:06 -0700 (0:00:00.024)       0:09:48.908 ***** 
> fatal: [/home/rmeggins/.cache/linux-system-roles/rhel-8-y.qcow2]: FAILED! => {
>     "changed": false,
>     "cmd": "/usr/sbin/subscription-manager register --org VALUE_SPECIFIED_IN_NO_LOG_PARAMETER --auto-attach --username VALUE_SPECIFIED_IN_NO_LOG_PARAMETER --password '********'",
>     "rc": 70
> }
> 
> STDOUT:
> 
> Registering to: localhost:8443/candlepin
> 
> STDERR:
> 
> Proxy error, unable to connect to proxy server.
> ```

Following #30, I changed `proxy_noauth_hostname` from "127.0.0.1" to "localhost"... and forgot to change the check in `setup_squid.yml:
```yaml
- name: Install and configure squid
  when:
    - lsr_rhc_test_data.proxy_noauth_hostname == "127.0.0.1"
```
:expressionless: :expressionless:

The fix is on the way.

ptoscano:
[citest]

ptoscano:
[citest bad]

defaults/main.yml
```diff
@@ -3,6 +3,7 @@
 rhc_auth: {}
 rhc_baseurl: null
 rhc_organization: null
+rhc_proxy: null
```
richm, line 6:
```suggestion
rhc_proxy: {}
```

README.md
```diff
@@ -98,6 +98,26 @@ rhc_repositories:
...
+```
+- `hostname` is the hostname of the proxy server
+- `port` is the port to which connect to on the proxy server
+- `username` is the username to use for authenticating on the proxy server;
+  it can be not specified if the proxy server does not require authentication
+- `password` is the password to use for authenticating on the proxy server;
+  it can be not specified if the proxy server does not require authentication
+
+Please note that specifying an empty dictionary, i.e. `{}`, *resets* all the
+proxy configurations to empty (effectively disabling the proxy server).
```
richm, line 119: hmm - this means that `rhc_proxy: {}` has a different semantic meaning than `rhc_auth: {}` and `rhc_server: {}`?  That seems inconsistent

richm: The preferred method of doing "specifying this parameter to this value will erase/reset configuration" is to use the `state: absent` thing, just like `rhc_release`.  So perhaps
```
If you want to erase the proxy settings, specify

    rhc_proxy:
      state: absent
```

ptoscano: oh, good idea, will switch to that approach

also note that `rhc_auth` and `rhc_proxy` are different anyway: you should be able to reset `rhc_proxy`, while "resetting" does not make sense for `rhc_auth`

ptoscano:
[citest]

richm (approved):

PR #29, opened 2022-12-07
Workaround redhat_subscription requiring creds even when unneeded

ptoscano:
The 'redhat_subscription' module currently requires parameters for credentials in case state=present. This applies also in case the system is already registered (and force_register=false), being then effectively unused. This is proposed as change to 'redhat_subscription' as: https://github.com/ansible-collections/community.general/pull/5664

In the meanwhile, workaround this situation locally in the role:
- add a fake credential string as private variable
- pass the fake credential string to both 'username' and 'password' in case rhc_state=present and the system is already registered

This way, all the users of the role do not need to specify credentials when not needed; this reflects in the tests, from which it is possible to drop the extra credentials.

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -31,8 +31,14 @@
 - name: Call subscription-manager
   community.general.redhat_subscription:
     state: "{{ rhc_state | d('present') }}"
-    username: "{{ rhc_auth.login.username | d(omit) }}"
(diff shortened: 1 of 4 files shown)


ptoscano:
[citest]

richm (approved):

ptoscano:
[citest]

PR #31, opened 2022-12-07
use tox-lsr 2.14.1

richm:


richm:
* tox-lsr moved to use ansible-core 2.14 which does not support python 3.8 - rather than try to revert that, I moved some of the ansible related checks to use python 3.9
* for some reason the latest ansible-test image does not work with docker - so install the `podman-docker` package to "fool" the system into using `podman` instead of `docker` - the package is only available on `ubuntu-latest` - in order to do this, we first have to remove `moby-cli` (aka `docker-client`)

This isn't pretty, and I have a big plan to replace all of this stuff with
* individual github actions for yamllint, ansible-lint, etc.
* removal of py, pylint, etc. tests for non-python roles

But for now this should get CI working again. If you guys approve, I'll roll this out to all roles.

PR #34, opened 2022-12-08
Fix passing 'server_insecure' to 'redhat_subscription'

ptoscano:
The role parameter 'rhc_server.insecure' is a boolean parameter, whereas the configuration value for subscription-manager is a string with the values '0' (== secure) or '1' (== insecure). Thus, passing the boolean value of the role parameter as-is to 'redhat_subscription' is not correct.

Use a slightly more complex approach for it:
- if 'rhc_server.insecure' is not passed or it was passed as "omit" value, then omit it
- if 'rhc_server.insecure' is true, pass '1'
- if 'rhc_server.insecure' is false, pass '0'


tasks/subscription-manager.yml
```diff
@@ -39,7 +39,14 @@
     server_hostname: "{{ rhc_server.hostname | d(omit) }}"
     server_port: "{{ rhc_server.port | d(omit) }}"
     server_prefix: "{{ rhc_server.prefix | d(omit) }}"
-    server_insecure: "{{ rhc_server.insecure | d(omit) }}"
+    server_insecure: "{% if ( 'insecure' not in rhc_server
```
richm, line 42:
```
"{{ omit if (rhc_server.insecure | d(omit) == omit)
     else rhc_server.insecure | ternary('1', '0') }}"
```

ptoscano:
[citest]

richm (approved):

PR #39, opened 2022-12-16
Add check for non-inclusive language

nhosoi:
Add a check for usage of terms and language that is considered
non-inclusive. We are using the woke tool for this with a wordlist
that can be found at
https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/woke.yml

tests/tasks/setup_candlepin.yml - cleanup non-inclusive words.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -24,6 +24,7 @@
     - name: Clone ansible-role-candlepin repo
       git:
         repo: https://github.com/candlepin/ansible-role-candlepin.git
+        # wokeignore:rule=master
         version: master
         dest: "{{ playbook_dir }}/roles/candlepin"
       delegate_to: 127.0.0.1
--- a/.github/workflows/woke.yml
+++ b/.github/workflows/woke.yml
@@ -1,0 +1,18 @@
+# yamllint disable rule:line-length
+name: Check for non-inclusive language
+on:  # yamllint disable-line rule:truthy
+  - pull_request
+jobs:
+  woke:
+    name: woke
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: woke
+        uses: get-woke/woke-action@v0
+        with:
+          woke-args: "-c https://raw.githubusercontent.com/linux-system-roles/tox-lsr/main/src/tox_lsr/config_files/woke.yml"
+          # Cause the check to fail on any broke rules
+          fail-on-error: true


tests/tasks/setup_candlepin.yml
```diff
@@ -22,9 +22,9 @@
         - ansible_distribution_major_version | int > 8
 
     - name: Clone ansible-role-candlepin repo
+      # noqa git-latest
       git:
         repo: https://github.com/candlepin/ansible-role-candlepin.git
-        version: master
```
ptoscano, line 27: while I understand the concern, this is a branch name, and as such it cannot be avoided...

since I'd like to keep the explicit specification of the branch checked out, is there any possibility to add a local override?

richm: > while I understand the concern, this is a branch name, and as such it cannot be avoided...
> 
> since I'd like to keep the explicit specification of the branch checked out, is there any possibility to add a local override?

If I understand how git clone works, if you do not specify a `version` (i.e. a branch name, tag, commit hash, etc.), then git clone will use the default branch configured in github.  This is what we would prefer.

If we must use `version: master` then we will need to use `version: master  # wokeignore:rule=master`

ptoscano: > If I understand how git clone works, if you do not specify a `version` (i.e. a branch name, tag, commit hash, etc.), then git clone will use the default branch configured in github.

This is true, and it works fine for my own personal interactive usage. Since here it is about automating, I think what should be done here is making sure to pull the wanted branch without relying on its default configuration.

> If we must use `version: master` then we will need to use `version: master # wokeignore:rule=master`

I'd personally lean toward this: not ideal, but still an identifier used by something else (a git repository).

nhosoi: > making sure to pull the wanted branch without relying on its default configuration

Yes, it is one advantage of specifying `version: master`. But the current trend is switching the primary branch name from `master` to `main`. It might also happen in ansible-role-candlepin. Then, if we are not aware of it, the test may start failing due to the branch `master` not being updated anymore... I might be wrong or am worried too much about something not needing to be, though.

ptoscano: > It might also happen in ansible-role-candlepin.

It's more likely that we will switch to some other solution, rather than that repository getting a branch rename.

nhosoi: ??? Sorry. Could you give me some more details? Do you mean with the "other solution", can we avoid having `version: master` in https://github.com/linux-system-roles/rhc?

ptoscano: It was more a long-term change to a different way to setup candlepin for testing, more like "thinking loud" -- nothing actionable right now.

Anyway: I'd rather keep branches explicitly specified, if possible -- if that has to be "master", so be it...

ptoscano (approved):
Thanks! LGTM now.

tests/tasks/setup_candlepin.yml
nhosoi: Thank you, @ptoscano. I've updated tests/tasks/setup_candlepin.yml as @richm described in https://github.com/linux-system-roles/rhc/pull/39#discussion_r1051442921.

PR #40, opened 2022-12-19
Implement insights "force_register" parameter

ahitacat:
Add a parameter for force registration of insights. This will unregister and register the system insights-client.

Extend the API with the  parameter:
-  `rhc_insights.force_register` this parameter is a boolean.

By default, this parameter is false. Only set to true to perform force registration.

Signed-off-by: Alba Hita Catala <ahitacat@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,14 +97,19 @@ rhc_repositories:
 
 A release to set for the system. Use `{"state":"absent"}` to actually unset the
 release set for the system.
-```
-rhc_insights:
-  state: present
-```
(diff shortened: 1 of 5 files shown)


ptoscano:
Hmm I have an almost finished implementation of `rhc_state: reconnect` that does this for both rhsm and insights. In particular, I think we should aim for a single state/variable to handle the forceful reconnection of both at once, and avoid playing with more combination of variables.

PR #41, opened 2023-01-02
ansible-lint 6.x fixes

richm:
Fix or suppress jinja spacing errors
All task names must begin with uppercase letter
All documents should begin with `---`
Use `true`/`false` instead of `yes`/`no`
The `block` should come last before all `when`, `vars`, etc.
Add `changed_when` for command, shell
Other cleanup


richm:
[citest]

ptoscano (approved):

PR #43, opened 2023-01-03
Implement `rhc_state: reconnect`

ptoscano:
Add an extra value to the `rhc_state` parameter to force the reconnection/registration of an already connected system, i.e. by disconnecting it first. When forcing the registration, some of the tasks/variables are used even if the system was previously connected (and thus its previous status does not matter).

In case we need to reconnect:
- avoid some checks on the status of subscription-manager or insights-client, as we will need to reconnect anyway, so unconditionally disconnect
- apply some of the logic/flows used for (not forced) connection

Make sure to test this:
- add helper scripts to get the system UUID from `subscription-manager`, and from `insights-client` (indirectly)
- extend `tests_register_unregister.yml`, and `tests_insights_client_register.yml` to get the rhsm/insights UUID after the connection, forcing a new connection, and checking that the UUID changed
- extend `tests_release.yml` to check that the release is properly changed also when forcing

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,11 @@ includes these collections/modules, you should have to do nothing.
     rhc_state: present
 
 Whether the system is connected to Red Hat; valid values are `present`
-(to ensure registration/connection), and `absent`.
+(to ensure registration/connection), `absent`, and `reconnect`.
+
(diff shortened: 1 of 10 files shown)


ptoscano:
[citest]

richm (approved):

PR #44, opened 2023-01-04
tests: specify "rhc_baseurl" where needed

ptoscano:
The role now may perform package checks in some cases, and because of that it needs to make sure that the package manager has usable repositories. Hence, pass the baseurl configured (if any, hence the "omit") for the tests as "rhc_baseurl" parameter in all the "rhc" invocations that are supposed to work (i.e. not fail for expected failing parameters).


ptoscano:
[citest]

richm (approved):

PR #47, opened 2023-01-05
tests: added test to register with invalid credentials

Archana-PandeyM:
Added a test to verify that registrations fails with invalid credentials when using rhc system role.


ptoscano:
[citest]

richm (approved):

PR #52, opened 2023-01-13
tests: fix enforced squid authentication

ptoscano:
Tweak the squid configuration to properly enforce authentication:
- disable the http_access bits that allow any localhost/localnet clients
- require authentication to connect

Also, tweak the test password for the proxy to be "proxypass" rather than "proxyuser" (which is the authentication username).

This affected only tests_proxy when run using a self-deployed Candlepin & squid; both the role and the test already worked fine, as testing against external servers (especially properly configured proxy servers) showed.


ptoscano:
[citest]

ptoscano:
[citest bad]

PR #54, opened 2023-01-16
Fix rhc_auth.activation_keys.keys

ptoscano:
It seems that ".keys" for a dictionary is actually a reserved keyword, and thus it cannot be used; hence, use the array notation to look it up.

Also, auto-attach cannot be used when using activation keys, so do not enable it in that case.

Make sure to test this:
- extend the test data by adding all the pools for the "admin" user to the "default_key" (existing and with no pools), so auto-attach works
- add a simple test of registration with activation keys to tests_register_unregister

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -44,8 +44,8 @@
       if (rhc_state | d('present') == 'present'
            and __rhc_subman_identity.rc == 0)
       else (rhc_auth.login.password | d(omit)) }}"
-    activationkey: "{{ rhc_auth.activation_keys.keys | join(',')
-      if (rhc_auth.activation_keys.keys is defined) else omit }}"
+    activationkey: "{{ rhc_auth.activation_keys['keys'] | join(',')
+      if ('keys' in rhc_auth.activation_keys | d({})) else omit }}"
     org_id: "{{ rhc_organization | d(omit) }}"
     server_hostname: "{{ rhc_server.hostname | d(omit) }}"
     server_port: "{{ rhc_server.port | d(omit) }}"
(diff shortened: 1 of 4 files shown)


ptoscano:
[citest]

tasks/subscription-manager.yml
```diff
@@ -53,7 +53,8 @@
     server_insecure: "{{ omit if (rhc_server.insecure | d(omit) == omit)
       else rhc_server.insecure | ternary('1', '0') }}"
     auto_attach: "{{ true
-      if (rhc_state | d('present') in ['present', 'reconnect']) else omit }}"
+      if (rhc_state | d('present') in ['present', 'reconnect']
+          and rhc_auth.activation_keys.keys is not defined) else omit }}"
```
richm, line 57:
```suggestion
          and not 'keys' in rhc_auth.activation_keys | d({})) else omit }}"
```

tasks/subscription-manager.yml
```diff
@@ -44,7 +44,7 @@
       if (rhc_state | d('present') == 'present'
            and __rhc_subman_identity.rc == 0)
       else (rhc_auth.login.password | d(omit)) }}"
-    activationkey: "{{ rhc_auth.activation_keys.keys | join(',')
+    activationkey: "{{ rhc_auth.activation_keys['keys'] | join(',')
       if (rhc_auth.activation_keys.keys is defined) else omit }}"
```
richm, line 48:
```suggestion
      if ('keys' in rhc_auth.activation_keys | d({})) else omit }}"
```

ptoscano:
[citest]

richm (approved):

PR #55, opened 2023-01-17
tests: ensure 'expect' is installed for Candlepin

ptoscano:
'expect' is currently needed by the Candlepin script that generates and signs the test packages, so make sure it is installed.

This fixes the generation of the test RPM packages on EL7.


ptoscano:
[citest]

ptoscano:
FYI: this had no impact simply because there is no attempt to install packages from the subscribed content in the tests.

richm (approved):

PR #57, opened 2023-01-24
Implement "rhc_insights.remediation"

ahitacat:
- Add a new parameter `rhc_insights.remediation` to enable insights remediation.  This depends on additional packages that need to be installed and to enable and start a service.
- Add a configuration test that enables and disables remediation.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,7 @@
 rhc_auth: {}
 rhc_baseurl: null
 rhc_insights:
+  remediation: present
   state: present
 rhc_organization: null
 rhc_proxy: {}
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -18,3 +18,27 @@
       or "NOT" in __rhc_insights_status.stdout
       or "unregistered" in __rhc_insights_status.stdout
   register: __rhc_insights_registration_result
+
+- name: Configure remediation
+  when: rhc_insights.remediation | d("present") == "present"
+  block:
+    - name: Ensure worker and ansible packages are installed
+      package:
+        name:
+          - rhc
+          - rhc-worker-playbook
+          - ansible-core
+        state: present
+
+    - name: Enable remediation
+      systemd:
+        enabled: true
+        state: started
+        name: rhcd
+
+- name: Disable remediation
+  systemd:
+    enabled: false
+    state: stopped
+    name: rhcd
+  when: rhc_insights.remediation | d("present") == "absent"
--- a/tests/tests_insights_remediation.yml
+++ b/tests/tests_insights_remediation.yml
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Test for insights remediation enabled & disabled
+  hosts: all
+  gather_facts: true
+
+  tasks:
+    - name: Skip when running with embedded Candlepin
+      meta: end_play
+      when:
+        - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Import test data
+      include_vars:
+        file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Register insights and enable remediation
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_auth:
+          login:
+            username: "{{ lsr_rhc_test_data.reg_username }}"
+            password: "{{ lsr_rhc_test_data.reg_password }}"
+        rhc_insights:
+          remediation: present
+          state: present
+        rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+        rhc_server:
+          hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+          port: "{{ lsr_rhc_test_data.candlepin_port }}"
+          prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+          insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+        rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+    - name: Get service_facts
+      service_facts:
+
+    - name: Check remediation is enabled
+      assert:
+        that:
+          - "'rhcd.service' in ansible_facts.services"
+          - ansible_facts.services['rhcd.service'].status == 'enabled'
+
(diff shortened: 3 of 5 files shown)


tasks/insights-client.yml
```diff
@@ -15,6 +15,30 @@
   when:
     - >-
       rhc_state | d("present") == "reconnect"
-      or "NOT" in __rhc_insights_status.stdout
+      "NOT" in __rhc_insights_status.stdout
```
ptoscano, line 18: careful, the "or" here is definitely needed!

tests/tests_insights_remediation.yml
```diff
@@ -0,0 +1,82 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Basic insights-client register/unregister
```
ptoscano, line 3: needs a more fitting description ;)

tests/tests_insights_remediation.yml
```diff
@@ -0,0 +1,82 @@
...
+        rhc_insights:
+          state: present
+          remediation: absent
+        rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+        rhc_server:
+          hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+          port: "{{ lsr_rhc_test_data.candlepin_port }}"
+          prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+          insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+        rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
```
ptoscano, line 57: you can remove all the vars but `rhc_insights.remediation`, as that's the only change we want to do here

tests/tests_insights_remediation.yml
```diff
@@ -0,0 +1,82 @@
...
```
ptoscano, line 76: ditto

tasks/main.yml
```diff
@@ -14,6 +14,8 @@
 
 - name: Handle system subscription
   include_tasks: subscription-manager.yml
+  tags:
+  - subscription-manager
```
ptoscano, line 18: uhm, what is it for?

defaults/main.yml
```diff
@@ -4,6 +4,7 @@ rhc_auth: {}
 rhc_baseurl: null
 rhc_insights:
   state: present
+  remediation: present
```
ptoscano, line 7: since there are other changes to do... a small nitpick here: can you please place it alphabetically sorted within `rhc_insights`? :)

ptoscano (changes requested):
I think it's a good start! There is only a critical issue (the removed "or"), and the rest are not too complicated notes/things to adjust.

Would it be difficult to check in the added test that the remediation was actually enabled (or disabled)? I think using `service_facts` should help.

tests/tests_insights_remediation.yml
richm: I suggest "Test for insights remediation register/unregister"

tasks/insights-client.yml
ahitacat: ouch I forgot to add it back

tasks/main.yml
ahitacat: Should be off, I use it to test Insights in a machine

ptoscano:
[citest]

ptoscano:
Thanks! I think it's nice now.

One more thing about the test: what about checking the status of the service? ie something like:
```yaml
    - service_facts:
    - assert:
        that:
          - "'rhcd.service' in ansible_facts.services"
          - ansible_facts.services['rhcd.service'].status == 'enabled'
```
right after the registration, and with `status == 'disabled'` after each role invocation that disables the remediation.

ahitacat:
> Thanks! I think it's nice now.
> 
> One more thing about the test: what about checking the status of the service? ie something like:
> 
> ```yaml
>     - service_facts:
>     - assert:
>         that:
>           - "'rhcd.service' in ansible_facts.services"
>           - ansible_facts.services['rhcd.service'].status == 'enabled'
> ```
> 
> right after the registration, and with `status == 'disabled'` after each role invocation that disables the remediation.

I thought about that. However, I'm using the systemd and package module, this test seems more like testing those modules work. But I can test it to check the conditionals are well set and working.

ptoscano:
[citest]

ptoscano (approved):
Since this cannot be tested in CI, I tested it manually and it seems to work nicely :+1: -> green approval mark

I'll wait for CI to complete to make sure there are no regressions in non-Insights code paths (hopefully not).

richm:
> > Thanks! I think it's nice now.
> > One more thing about the test: what about checking the status of the service? ie something like:
> > ```yaml
> >     - service_facts:
> >     - assert:
> >         that:
> >           - "'rhcd.service' in ansible_facts.services"
> >           - ansible_facts.services['rhcd.service'].status == 'enabled'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > right after the registration, and with `status == 'disabled'` after each role invocation that disables the remediation.
> 
> I thought about that. However, I'm using the systemd and package module, this test seems more like testing those modules work. But I can test it to check the conditionals are well set and working.

Well, yes and no.  I'm assuming the rhcd package provides the systemd configuration and /etc/sysconfig config file?  If so, then checking the status of the rhcd service is a good way to check
* the rhcd package was installed correctly
* the way the role configures rhcd is correct
* if there is an error further down the chain, you won't have to dig into the logs to figure out that the root of the problem is that the service wasn't started for some reason

richm:
The failure is coming from this task
```
TASK [Clone ansible-role-candlepin repo] ***************************************
task path: /WORKDIR/dist-git-rhc-feat-insights-remediation-iu29nz6n/tests/tasks/setup_candlepin.yml:25
Wednesday 25 January 2023  18:37:06 +0000 (0:00:01.194)       0:00:12.224 ***** 
fatal: [sut -> 127.0.0.1]: FAILED! => {
    "changed": false,
    "cmd": "/usr/bin/git clone --origin origin https://github.com/candlepin/ansible-role-candlepin.git /WORKDIR/dist-git-rhc-feat-insights-remediation-iu29nz6n/tests/roles/candlepin",
    "rc": 128
}

STDERR:

fatal: destination path '/WORKDIR/dist-git-rhc-feat-insights-remediation-iu29nz6n/tests/roles/candlepin' already exists and is not an empty directory.
```
I think we can merge this PR anyway

But this task should be changed to either use an existing candlepin directory if found, or each invocation should create and clone its own temp directory

ptoscano:
[citest bad]

PR #62, opened 2023-02-02
Implemented "rhc_insights.tags" parameter

ahitacat:
Add a new parameter to configure insights tags that will be added to the system record in HBI.

The syntax of this parameter is a dictionary, and the value for each key can be any data type.

The role will try to configure the tags if this parameter is present. If it is not present the tags will remain in the same status.

Test added that verifies:
- Add and delete tags
- Remove all tags
- Do not modify tags if the parameter is missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,29 @@ Whether the system is connected to Insights; valid values are `present`
 Whether the system is configured to run Insights remediation; valid values are
 `present` (to ensure remediation) and `absent`.
 
+    rhc_insights:
+      tags: {}
+
+A dictionary of tags that is added to the system record in Host Based Inventory (HBI).
+Used to search for hosts. This tags has keys and values. Values for each key can be any 
+data type. Posible values of `tags` parameter:
+- `tags` is `null` or an empty value (e.g: `{}`) does not change the tags file content.
+- `tags` is `{state: absent}` remove all the tags by removing the file.
+- `tags` is non-empty, create the file with the specified tags
+
+An example of the tags configured in the `insights-client` [documentation](https://access.redhat.com/documentation/en-us/red_hat_insights/2022/html/client_configuration_guide_for_red_hat_insights/con-insights-client-tagging-overview_insights-cg-adding-tags):
+
+```yaml
+rhc_insights:
+  tags:
+    group: _group-name-value_
+    location: _location-name-value_
+    description:
+      - RHEL8
+      - SAP
+    key 4: value
+```
+
 
     rhc_proxy: {}
 
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -10,6 +10,25 @@
   when:
     - rhc_state | d("present") != "reconnect"
 
+- name: Add insights tags
+  copy:
+    dest: /etc/insights-client/tags.yml
+    content: "{{ rhc_insights.tags | to_nice_yaml(indent=2) }}"
+    mode: '0644'
+  register: __insights_tags_added
+  when:
+    - rhc_insights.tags is defined
+    - rhc_insights.tags != __rhc_state_absent
+    - rhc_insights.tags | d({}) | length > 0
+
+- name: Delete insights tags
+  file:
+    state: absent
+    path: /etc/insights-client/tags.yml
+  register: __insights_tags_removed
+  when:
+    - rhc_insights.tags | d(None) == __rhc_state_absent
+
 - name: Register insights-client
   command: insights-client --register
   when:
@@ -18,6 +37,16 @@
       or "NOT" in __rhc_insights_status.stdout
       or "unregistered" in __rhc_insights_status.stdout
   register: __rhc_insights_registration_result
+
+- name: Do the collection if insights-client was registered
+  command: insights-client
+  when:
+    - >-
+      "This host is registered" in __rhc_insights_status.stdout
+      or "Registered" in __rhc_insights_status.stdout
+    - >-
+      __insights_tags_added is changed
+      or __insights_tags_removed is changed
 
 - name: Configure remediation
   when: rhc_insights.remediation | d("present") == "present"
--- a/tests/tasks/get_insights_tags.yml
+++ b/tests/tasks/get_insights_tags.yml
@@ -1,0 +1,6 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Get state of insights tag file
+  stat:
+    path: /etc/insights-client/tags.yml
+  register: test_insights_tags
--- a/tests/tests_insights_tags.yml
+++ b/tests/tests_insights_tags.yml
@@ -1,0 +1,106 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Test for insights tags
+  hosts: all
+  gather_facts: true
+
+  tasks:
+    - name: Skip when running with embedded Candlepin
+      meta: end_play
+      when:
+        - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Import test data
+      include_vars:
+        file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Configure tags and register insights
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_auth:
+          login:
+            username: "{{ lsr_rhc_test_data.reg_username }}"
+            password: "{{ lsr_rhc_test_data.reg_password }}"
+        rhc_insights:
+          state: present
+          tags:
+            group: ansible
+            activity: test
+        rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+        rhc_server:
+          hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+          port: "{{ lsr_rhc_test_data.candlepin_port }}"
+          prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+          insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+        rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+    - name: Get state of insights tags.yml file
+      include_tasks: get_insights_tags.yml
+
+    - name: Rename the tags to test_insights_tags_modified
+      set_fact:
+        test_insights_tags_new: "{{ test_insights_tags }}"
+
+    - name: Check if tags.yml file exists and its size is > 0
+      assert:
+        that: test_insights_tags_new.stat.size > 0
+
+    - name: Change tags
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          tags:
+            group:
+              - ansible
+              - rhc
+
+    - name: Get state of insights tags.yml file
+      include_tasks: get_insights_tags.yml
+
+    - name: Rename the tags to test_insights_tags_modified
+      set_fact:
+        test_insights_tags_modified: "{{ test_insights_tags.stat.size }}"
+
+    - name: Check if tags.yml file exists and its size has changed
+      assert:
+        that:
+          - test_insights_tags_new.stat.size != test_insights_tags_modified
+
+    - name: Do nothing
+      include_role:
+        name: linux-system-roles.rhc
+
+    - name: Get state of insights tags.yml file
+      include_tasks: get_insights_tags.yml
+
+    - name: Check if tags.yml file exists and its size is the same
+      assert:
+        that:
+          - test_insights_tags.stat.size == test_insights_tags_modified
+
+    - name: Remove all tags
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          tags:
+            state: absent
+
+    - name: Get state of insights tags.yml file
+      include_tasks: get_insights_tags.yml
+
+    - name: Check that insights tags.yml file does not exists
+      assert:
+        that: test_insights_tags is undefined
+
+    - name: Unregister
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_state: absent
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,6 +6,7 @@ rhc_environments: []
 rhc_insights:
   remediation: present
   state: present
+  tags: {}
 rhc_organization: null
 rhc_proxy: {}
 rhc_release: null


tasks/insights-client.yml
```diff
@@ -10,6 +10,13 @@
   when:
     - rhc_state | d("present") != "reconnect"
 
+- name: Configure tags
+  copy:
+    dest: /etc/insights-client/tags.yml
+    content: "{{ rhc_insights.tags | to_nice_yaml }}"
+    mode: '0644'
+  when: rhc_insights.tags is defined
```
ptoscano, line 18: i think that, since `tags` was added in `defaults/main.yml`, this check will be always true

ptoscano (changes requested):
It looks like a good starting point! I added an inline note.

General notes/questions:
- it definitely needs to be documented in `README.md` :)
- in case the tags are modified on an already connected system, is there anything to trigger for the new tags to take effect?
- what if the user wants to remove all the tags? A possibility here could be to do something similar to what `rhc_release` and `rhc_proxy` do to unset things: in case the value is the special `{"state":"absent"}` (see `__rhc_state_absent`), remove the tags file altogether.
- don't forget to unsubscribe the system at the end of the test
- maybe add minimal checks that `/etc/insights-client/tags.yml` exists or not depending on the steps in the test to check that something is there (or not)

I'll try to give it a basic testing tomorrow.

tasks/insights-client.yml
richm: > i think that, since `tags` was added in `defaults/main.yml`, this check will be always true

It might not be defined - consider if someone uses `rhc_insights: {}`
this should work `rhc_insights.tags | d({}) | length > 0` - an empty `dict` should have a length of 0, otherwise, the length should be the number of keys

richm:
> It looks like a good starting point! I added an inline note.
> 
> General notes/questions:
> 
>     * it definitely needs to be documented in `README.md` :)

+1
 
>     * in case the tags are modified on an already connected system, is there anything to trigger for the new tags to take effect?

If so, use a `register: __insights_tags` on the `copy` task, then you can trigger something by using `when: __insights_tags is changed` (or possibly use a handler)

> 
>     * what if the user wants to remove all the tags? A possibility here could be to do something similar to what `rhc_release` and `rhc_proxy` do to unset things: in case the value is the special `{"state":"absent"}` (see `__rhc_state_absent`), remove the tags file altogether.

+1
 
>     * don't forget to unsubscribe the system at the end of the test
> 
>     * maybe add minimal checks that `/etc/insights-client/tags.yml` exists or not depending on the steps in the test to check that something is there (or not)

+1 - use the `stat` module https://docs.ansible.com/ansible/2.9/modules/stat_module.html#stat-module - see if the file exists (or not if removing it) and see if the `size` is greater than 0

> 
> 
> I'll try to give it a basic testing tomorrow.

ahitacat:
> * it definitely needs to be documented in `README.md` :)
I forgot about this.

> * in case the tags are modified on an already connected system, is there anything to trigger for the new tags to take effect?

I was thinking that this could be updated in the next scheduled collection. But I do not see any problem to do a collection after the tags changing if insights-client was already registered.

> * what if the user wants to remove all the tags? A possibility here could be to do something similar to what `rhc_release` and `rhc_proxy` do to unset things: in case the value is the special `{"state":"absent"}` (see `__rhc_state_absent`), remove the tags file altogether.

This was done with the conditional `rhc_insights.tags is defined`, but @richm suggested a better approach.

> * don't forget to unsubscribe the system at the end of the test

:100:  I forgot about this, as well.

> * maybe add minimal checks that `/etc/insights-client/tags.yml` exists or not depending on the steps in the test to check that something is there (or not)

Added those checks, using `stat` for this purpose.

I have pushed new changes to address all these comments.

README.md
```diff
@@ -116,6 +116,23 @@ Whether the system is connected to Insights; valid values are `present`
 Whether the system is configured to run Insights remediation; valid values are
 `present` (to ensure remediation) and `absent`.
 
+    rhc_insights:
+      tags: {}
+
+A dictionary of tags that is added to the system record in Host Based Inventory (HBI).
+Used to search for hosts. This tags has keys and values. Values for each key can be any 
+data type. An example of the tags configured in the `insights-client` [documentation](https://access.redhat.com/documentation/en-us/red_hat_insights/2022/html/client_configuration_guide_for_red_hat_insights/con-insights-client-tagging-overview_insights-cg-adding-tags):
```
richm, line 124: If `tags` is `null` or empty, nothing is done.  Existing tags are replaced with the given tags.  If you want to remove all tags and the tags.yaml file, use `tags: {state: "absent"}`

README.md
```diff
@@ -116,6 +116,23 @@ Whether the system is connected to Insights; valid values are `present`
...
+
+```yaml
+rhc_insights:
+  tags:
+    group: _group-name-value_
+    location: _location-name-value_
+    description:
+    - RHEL8
```
richm, line 132:
```suggestion
      - RHEL8
```

README.md
```diff
@@ -116,6 +116,23 @@ Whether the system is connected to Insights; valid values are `present`
...
+    - SAP
```
richm, line 133:
```suggestion
      - SAP
```

tasks/insights-client.yml
```diff
@@ -10,15 +10,41 @@
   when:
     - rhc_state | d("present") != "reconnect"
 
+- name: Add insights tags
+  copy:
+    dest: /etc/insights-client/tags.yml
+    content: "{{ rhc_insights.tags | to_nice_yaml }}"
```
richm, line 16: If you need to, you can change the formatting and indentation of the generated yaml - see https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#formatting-data-yaml-and-json and https://jinja.palletsprojects.com/en/2.9.x/templates/#indent

tasks/insights-client.yml
```diff
@@ -10,15 +10,41 @@
...
+  when: rhc_insights.tags | d({}) | length == 0
+
 - name: Register insights-client
   command: insights-client --register
   when:
     - >-
       rhc_state | d("present") == "reconnect"
       or "NOT" in __rhc_insights_status.stdout
       or "unregistered" in __rhc_insights_status.stdout
+
```
richm, line 35:
```suggestion
```

README.md
ptoscano: Hmm if empty means "no changes" too, then there is no way for the user to remove all the tags. I think a better way could be:
- `tags` is `null` -> no changes
- `tags` is `{}` -> remove all the tags (removing the file)
- `tags` is non-empty -> create/replace the file with the specified tags

tests/tasks/get_insights_tags.yml
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Get state of insights tag file
+  stat:
+    path: /etc/insights-client/tags.yml
+  register: __tags_file
```
ptoscano, line 6: I think you can skip the extra variable, and directly register `test_insights_tags` (only a minor thing)

tests/tests_insights_tags.yml
```diff
@@ -0,0 +1,110 @@
...
+    - name: Check if tags.yml file exists and its size is the same
+      assert:
+        that:
+          - test_insights_tags.stat.size == test_insights_tags_modified
+
+    - name: Remove all tags
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
```
ptoscano, line 97: this will probably require to specify `tags` somehow, as not specifying it means it is `null`, and thus "no changes"

tests/tests_insights_tags.yml
```diff
@@ -0,0 +1,110 @@
...
+
+    - name: Get state of insights tags.yml file
+      include_tasks: get_insights_tags.yml
+
+    - name: Check that insights tags.yml file does not exists
+      assert:
+        that: test_insights_tags.stat.exists == False
```
ptoscano, line 104: or simply "not var" ;)

README.md
ahitacat: I would try to avoid to use the content of tags to check if the file needs to be removed. Although it is not a common tag there are not restrictions to setting those, so I would like to no add restriction here neither.

@ptoscano proposition is what I wanted. 
tags default value should be None in this case.

tasks/insights-client.yml
ahitacat: I think is not needed to change to much the formatting, it is a simple yaml file.

tests/tasks/get_insights_tags.yml
ahitacat: You think right :smile:

tests/tests_insights_tags.yml
ahitacat: :100:

tests/tests_insights_tags.yml
ahitacat: I have tried this, but this var is registered so it doesn't work. I'll keep this.

tasks/insights-client.yml
```diff
@@ -10,6 +10,25 @@
...
+    - rhc_insights.tags | d({}) | length > 0
+
+- name: Delete insights tags
+  file:
+    state: absent
+    path: /etc/insights-client/tags.yml
+  register: __insights_tags_removed
+  when:
+    - rhc_insights.tags != None
+    - rhc_insights.tags | d({}) | length == 0
```
richm, line 30: This is a little bit tricky - you want to remove the file if, and only if, `rhc_insights.tags == {}` - so perhaps
`when: rhc_insights.tags | d(None) == {}`
in this case `rhc_insights.tags` must be explicitly defined to the value `{}` - otherwise, the file is not removed - otherwise, cases like `rhc_insights.tags: []` or `rhc_insights.tags: ""` will cause the file to be removed since the length is 0.

In general, it is better to be explicit than implicit.

tests/tests_insights_tags.yml
```diff
@@ -0,0 +1,106 @@
...
+
+    - name: Check if tags.yml file exists and its size has changed
+      assert:
+        that:
+          - test_insights_tags_new.stat.size != test_insights_tags_modified
+
+    - name: Do nothing
+      include_role:
+        name: linux-system-roles.rhc
+
```
richm, line 78:
```suggestion
```

README.md
```diff
@@ -116,6 +116,29 @@ Whether the system is connected to Insights; valid values are `present`
...
 
+    rhc_insights:
+      tags: None
+
+A dictionary of tags that is added to the system record in Host Based Inventory (HBI).
+Used to search for hosts. This tags has keys and values. Values for each key can be any 
+data type. Posible values of `tags` parameter:
+- `tags` is null does not change the tags file content.
+- `tags` is {} remove all the tags by removing the file.
+- `tags` is non-empty, create the file with the specified tags
```
richm, line 127: The problem with this is that in other similar cases in this role, the use of `state: absent` is used to remove something.  Now the user will have to remember "In case of registration and proxy, I have to use `state: absent`, but in the case of tags, I have to use `{}` to do the same thing."

ptoscano: Oops, you are right, and I even mentioned `{"state":"absent"}` in my first comment... :-(

Alba: sorry for the wrong suggestion today, let's adopt the same style of `{"state":"absent"}` for removing tags.

ahitacat: Ok, this makes the role more robust.

tasks/insights-client.yml
```diff
@@ -10,6 +10,26 @@
...
+    - rhc_insights.tags is defined
+    - rhc_insights.tags.state | d(omit) != "absent"
+    - rhc_insights.tags | d({}) | length > 0
+
+- name: Delete insights tags
+  file:
+    state: absent
+    path: /etc/insights-client/tags.yml
+  register: __insights_tags_removed
+  when:
```
richm, line 29:
```suggestion
  when: rhc_insights.tags | d(None) == {"state": "absent"}
```
I think that's the only condition you need - only if rhc_insights.tags is defined and equal to `{"state": "absent"}` will the tags be removed and the file deleted

ptoscano: Or, instead of hardcoding `{"state": "absent"}`, use the already-made `__rhc_state_absent` variable ;)

README.md
```diff
@@ -116,6 +116,29 @@ Whether the system is connected to Insights; valid values are `present`
...
 `present` (to ensure remediation) and `absent`.
 
+    rhc_insights:
+      tags: {}
+
+A dictionary of tags that is added to the system record in Host Based Inventory (HBI).
+Used to search for hosts. This tags has keys and values. Values for each key can be any 
+data type. Posible values of `tags` parameter:
+- `tags` is null or {} does not change the tags file content.
+- `tags` is {state: absent} remove all the tags by removing the file.
```
ptoscano, line 126: minor: use `-quoting for `{state: absent}`

README.md
```diff
@@ -116,6 +116,29 @@ Whether the system is connected to Insights; valid values are `present`
 Whether the system is configured to run Insights remediation; valid values are
 `present` (to ensure remediation) and `absent`.
 
+    rhc_insights:
+      tags: {}
+
+A dictionary of tags that is added to the system record in Host Based Inventory (HBI).
+Used to search for hosts. This tags has keys and values. Values for each key can be any 
+data type. Posible values of `tags` parameter:
+- `tags` is null or {} does not change the tags file content.
```
ptoscano, line 125: minor: use `-quoting for `{}`

richm, line 125:
```suggestion
- `tags` is `null` or `{}` does not change the tags file content.
```

README.md
richm, line 126:
```suggestion
- `tags` is `{state: absent}` remove all the tags by removing the file.
```

tasks/insights-client.yml
```diff
@@ -10,6 +10,25 @@
...
 
+- name: Add insights tags
+  copy:
+    dest: /etc/insights-client/tags.yml
+    content: "{{ rhc_insights.tags | to_nice_yaml(indent=2) }}"
+    mode: '0644'
+  register: __insights_tags_added
+  when:
+    - rhc_insights.tags is defined
+    - rhc_insights.tags.state | d(omit) != "absent"
```
richm, line 21:
```suggestion
    - rhc_insights.tags != __rhc_state_absent
```

richm (changes requested):
Please make the suggested change at tasks/insights-client.yml line 21

ahitacat:
Sorry @richm I didn't saw that line

richm:
[citest]

richm:
baseos ci is acting up :-(
[citest]

richm:
lgtm - waiting for @ptoscano to clear the "change requested" - also, we need to either solve these candlepin installation issues on EL7, or skip candlepin installation on EL7

ptoscano (approved):
> lgtm - waiting for @ptoscano to clear the "change requested"

LGTM too, thanks!

> also, we need to either solve these candlepin installation issues on EL7, or skip candlepin installation on EL7

IIRC since yesterday candlepin/master does not support EL7 anymore... OTOH, in internal discussions we decided to not support the role on EL7, due to where it is in its lifecycle. I'll drop it from `meta/main.yml` soon.

PR #66, opened 2023-02-09
meta: stop supporting EL7

ptoscano:
After internal discussions, we decided to not support EL7 due to where it is in its lifecycle.


ptoscano:
[citest]

richm (approved):

PR #67, opened 2023-02-09
Added "rhc_insights.autoupdate" parameter

ahitacat:
Add a new parameter to configure insights auto_update parameter. This will moified the insights-client configuration file.

The syntax of this parameter is a boolean.

Add test for verification.

### DIFF
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -29,6 +29,13 @@
   when:
     - rhc_insights.tags | d(None) == __rhc_state_absent
 
+- name: Configure autoupdate
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^auto_update"
+    insertafter: "#auto_update"
+    line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
+
 - name: Register insights-client
   command: insights-client --register
   when:
--- a/tests/tests_insights_autoupdate.yml
+++ b/tests/tests_insights_autoupdate.yml
@@ -1,0 +1,65 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Basic insights autoupdate test
+  hosts: all
+  gather_facts: true
+  become: true
+
+  tasks:
+    - name: Skip when running with embedded Candlepin
+      meta: end_play
+      when:
+        - lookup('env', 'LSR_RHC_TEST_DATA') | length == 0
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Import test data
+      include_vars:
+        file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Test autoupdate
+      block:
+        - name: Add autoupdate and register insights
+          include_role:
+            name: linux-system-roles.rhc
+            public: true
+          vars:
+            rhc_insights:
+              autoupdate: true
+              remediation: absent
+              state: present
+
+        - name: Check that auto_update is true in config file
+          command:
+            grep -ixq "^auto_update=true" {{ __rhc_insights_conf }}
+          changed_when: false
+
+        - name: Disable autoupdate
+          include_role:
+            name: linux-system-roles.rhc
+          vars:
+            rhc_insights:
+              autoupdate: false
+              remediation: absent
+
+        - name: Disable autoupdate (noop)
+          include_role:
+            name: linux-system-roles.rhc
+          vars:
+            rhc_insights:
+              autoupdate: false
+              remediation: absent
+
+        - name: Check that auto_update is false in config file
+          command:
+            grep -ixq "^auto_update=false" {{ __rhc_insights_conf }}
+          changed_when: false
+
+      always:
+        - name: Unregister
+          include_role:
+            name: linux-system-roles.rhc
+          vars:
+            rhc_state: absent
--- a/README.md
+++ b/README.md
@@ -111,6 +111,12 @@ Whether the system is connected to Insights; valid values are `present`
 
 
     rhc_insights:
+      autoupdate: true
+
+Whether the system automatically updates the dynamic configuration. By default is
+`true`.
+
+    rhc_insights:
       remediation: present
 
 Whether the system is configured to run Insights remediation; valid values are
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -4,6 +4,7 @@ rhc_auth: {}
 rhc_baseurl: null
 rhc_environments: []
 rhc_insights:
+  autoupdate: true
   remediation: present
   state: present
   tags: {}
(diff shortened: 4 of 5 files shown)


tasks/insights-client.yml
```diff
@@ -29,6 +29,13 @@
   when:
     - rhc_insights.tags | d(None) == __rhc_state_absent
 
+- name: Configure autoupdate
+  lineinfile:
+    path: /etc/insights-client/insights-client.conf
+    regexp: "^auto_update"
+    insertafter: "#auto_update"
+    line: "auto_update={{ rhc_insights.autoupdate| d(omit) | bool }}"
```
richm, line 37:
```suggestion
    line: auto_update={{ rhc_insights.autoupdate | d(omit) | bool }}
```
This seems a little bit tricky.  There are the following cases:
* user does not define `rhc_insights` in their inventory/playbook - then the default value will be used, which means `rhc_insights.autoupdate: true`
* user sets `rhc_insights: {}` or some other empty value - in this case, we do not want to touch insights-client.conf
* user sets `rhc_insights` and sets parameters other than `autoupdate` e.g. `rhc_insights.remediation` - in that case we want to use the default value of `rhc_insights.autoupdate: true`
* user explicitly sets `rhc_insights.autoupdate: $value`

note that if the user does something like `rhc_insights.autoupdate: none` or some value that is not `true` or `false`, we should use the `| bool` filter to attempt to convert the value to a boolean, and let the role fail if the value cannot be converted to `true` or `false`

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,68 @@
...
+      vars:
+        rhc_insights:
+          autoupdate: true
+          state: present
+
+    - name: Get the auto_update line in file configuration
+      command:
+        grep -ixq "^auto_update=true" /etc/insights-client/insights-client.conf
+      register: test_autoupdate_conf
+      ignore_errors: yes
```
richm, line 34:
```suggestion
      failed_when: false
      changed_when: false
```

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    - name: Check if auto_update is enabled in config file
+      assert:
+        that: test_autoupdate_conf.rc == 0
```
richm, line 38: You don't really need this check here - you could omit the `ignore_errors` and `failed_when` in the `Get the auto_update line in file configuration` task - if that task fails then the test will report an error there

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    - name: Disable autoupdate
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          autoupdate: false
+
```
richm, line 46:
```suggestion

```

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,68 @@
...
+    - name: Get the auto_update line in file configuration
+      command:
+        grep -ixq "^auto_update=false" /etc/insights-client/insights-client.conf
+      register: test_autoupdate_conf
+      ignore_errors: yes
+
+    - name: Check if auto_update is disabled in config file
+      assert:
+        that: test_autoupdate_conf.rc == 0
```
richm, line 62: see above for the similar assert task

tasks/insights-client.yml
ahitacat: Yes it seems: I opted to change the `omit` to `true`. That way the playbook assures that is `true` by default, whether is not present or explicit expressed. (Which is also the default value of `insights-client`)

tests/tests_insights_autoupdate.yml
ahitacat: you are right!

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,58 @@
...
+      changed_when: false
+
+    - name: Disable autoupdate
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          autoupdate: false
+
+    - name: Disable autoupdate
```
ptoscano, line 42:
```suggestion
    - name: Disable autoupdate (noop)
```

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    - name: Add autoupdate and register insights
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          autoupdate: true
+          state: present
+
+    - name: Get the auto_update line in file configuration
```
ptoscano, line 30:
```suggestion
    - name: Check that auto_update is true
```

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,58 @@
...
+          autoupdate: false
+
+    - name: Disable autoupdate
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          autoupdate: false
+
+    - name: Get the auto_update line in file configuration
```
ptoscano, line 49:
```suggestion
    - name: Check that auto_update is false
```

ptoscano (commented):
Few notes on the test:

- it should register first: add the registration step, and put all the commands in a `block`, with the unregistration in the `always` part of the `block`
- I'd set `rhc_insights.remediation: absent` in all the tasks, so there are less things done during the test

tasks/insights-client.yml
```diff
@@ -29,6 +29,13 @@
   when:
     - rhc_insights.tags | d(None) == __rhc_state_absent
 
+- name: Configure autoupdate
+  lineinfile:
+    path: /etc/insights-client/insights-client.conf
```
richm, line 34:
```suggestion
    path: "{{ __rhc_insights_conf }}"
```

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,64 @@
...
+      include_vars:
+        file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+      delegate_to: 127.0.0.1
+      become: false
+
+    - name: Test autoupdate
+      block:
+        - name: Add autoupdate and register insights
+          include_role:
+            name: linux-system-roles.rhc
```
richm, line 26:
```suggestion
            name: linux-system-roles.rhc
            public: true
```
Otherwise, I don't think the `__rhc_insights_conf` variable will be exported from the role

richm: you don't need to add `public: true` below - once it is exported, it is available for the rest of the playbook

tests/tests_insights_autoupdate.yml
```diff
@@ -0,0 +1,64 @@
...
+          vars:
```
richm, line 27:
```suggestion
            public: true
          vars:
```

richm:
[citest]

richm (approved):

PR #68, opened 2023-02-10
Improve tests for insights tags

ahitacat:
Added a block and always section to unregister the system in case the test fails.

Disable remediation from the playbook to reduce the tasks done.


richm:
[citest]

richm:
[citest]

PR #70, opened 2023-02-14
Fix rhc_insights.remediation when absent

ahitacat:
This tasks needs to check if the package is installed in the system before trying to disable it via systemd.

Added rhc_insights.remediation absent to one task of tags test that was missed to configure.


richm:
[citest]

richm (approved):

PR #75, opened 2023-03-14
tests: refactor Candlepin setup

ptoscano:
While the existing `setup_candlepin.yml` helper generally works fine, there are few drawbacks:
- in case we do not want to deploy Candlepin, the logic for reading the test data is duplicated in the tests
- in case a test is skipped because of a configuration in the test data of the deployed Candlepin, Candlepin will be deployed anyway
- because the test data of the deployed Candlepin is set after the deployment, all the data are manually set all over the place

To improve the situation, massage `setup_candlepin.yml` a bit:
- split the configuration of the deployed Candlepin in its own file (`candlepin_data.yml`)
- create `setup_test_data.yml`: this has the logic for importing the test data from an external file, or from `candlepin_data.yml`
- create `check_candlepin.yml`: small helper to check that the configured Candlepin set in the test data
- `setup_candlepin.yml` imports `setup_test_data.yml` at the beginning, and `check_candlepin.yml` at the end to retain its behaviour
- `setup_candlepin.yml` can now use the configuration of the test data, as it is set up before it deploys Candlepin

### DIFF
--- a/tests/tasks/setup_test_data.yml
+++ b/tests/tasks/setup_test_data.yml
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Get LSR_RHC_TEST_DATA environment variable
+  set_fact:
+    lsr_rhc_test_data_file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+  become: false
+
+- name: Import test data
+  include_vars:
+    file: "{{ lsr_rhc_test_data_file }}"
+  when: lsr_rhc_test_data_file | length > 0
+  become: false
+
+- name: Set local lsr_rhc_test_data
+  include_vars:
+    file: ../files/candlepin_data.yml
+  when: lsr_rhc_test_data_file | length == 0
--- a/tests/files/candlepin_data.yml
+++ b/tests/files/candlepin_data.yml
@@ -1,0 +1,32 @@
+# SPDX-License-Identifier: MIT
+---
+lsr_rhc_test_data:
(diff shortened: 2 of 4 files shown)


ptoscano:
[citest]

ptoscano:
[citest bad]

ptoscano:
Ah, I forgot to explicitly write it in the commit message (so writing it here, at least): this refactoring supposedly has no actual behaviour changes on how the tests work.

tests/tasks/setup_test_data.yml
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Get LSR_RHC_TEST_DATA environment variable
+  set_fact:
+    lsr_rhc_test_data_file: "{{ lookup('env', 'LSR_RHC_TEST_DATA') }}"
+  delegate_to: 127.0.0.1
```
richm, line 6: I don't think you need to delegate here - lookups work against the control node

tests/tasks/setup_test_data.yml
```diff
@@ -0,0 +1,19 @@
...
+  become: false
+
+- name: Import test data
+  include_vars:
+    file: "{{ lsr_rhc_test_data_file }}"
+  when: lsr_rhc_test_data_file | length > 0
+  delegate_to: 127.0.0.1
```
richm, line 13: I don't think you need to delegate here - include_vars should work against the control node

ptoscano:
[citest]

richm (approved):

PR #77, opened 2023-03-14
workaround insights-client issue with /usr/bin/python

richm:
insights-client does not work properly when it is spawned from
/usr/bin/python, but it does work when spawned from /usr/bin/python3,
even if they are symlinks to each other i.e. the same binary.
Not sure why this is, but the workaround is to start
insights-client in the background and `wait` for it.

### DIFF
--- a/tasks/check-insights-status.yml
+++ b/tasks/check-insights-status.yml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: MIT
 ---
 - name: Call insights status
-  command: insights-client --status
+  shell: insights-client --status & wait
   changed_when: false
   register: __rhc_insights_status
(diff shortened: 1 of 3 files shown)


richm:
@ahitacat fyi

richm:
[citest]

ahitacat:
This looks good to me, I have checked it with the `tox-lsr` for ansible-2.9 and the workaround seems to work.

```
  qemu-ansible-2.9: commands succeeded
  congratulations :)
```

ahitacat (approved):

ptoscano (approved):
A quick check passed, and @ahitacat 's testing was successful too -> approve.

PR #78, opened 2023-03-15
tests: do not deploy Candlepin in tests_release

ptoscano:
A local Candlepin deployment has no concept of "releases", so deploying it to unconditionally end the test is suboptimal.

Make use of the new helper tasks for the tests:
- only setup the test data first, so we get the configured release version for the test
- keep skipping the test if there is no configured release: this way, even an external Candlepin configuration may still skip the test
- ensure that the configured Candlepin works: `setup_candlepin.yml` does it automatically, so we need to manually use `check_candlepin.yml` now

The only behaviour change is that this test will end much faster when using the local Candlepin deployment (i.e. the default setup).

### DIFF
--- a/tests/tests_release.yml
+++ b/tests/tests_release.yml
@@ -7,13 +7,18 @@
   tags:
     - tests::slow
   tasks:
-    - name: Setup Candlepin
-      import_tasks: tasks/setup_candlepin.yml
+    # A local Candlepin deployment has no concept of "releases",
+    # so only setup the test data.
+    - name: Setup test data
+      import_tasks: tasks/setup_test_data.yml
 
     - name: Skip if no test release is set
       meta: end_play
       when:
         - lsr_rhc_test_data.release | d(none) is none
+
+    - name: Check Candlepin works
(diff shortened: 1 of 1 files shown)


tests/tests_release.yml
```diff
@@ -7,14 +7,19 @@
   tags:
     - tests::slow
   tasks:
-    - name: Setup Candlepin
-      import_tasks: tasks/setup_candlepin.yml
+    # A local Candlepin deployment has no concept of "releases",
+    # so onlt setup the test data
```
richm, line 11:
```suggestion
    # so only setup the test data
```

ptoscano:
[citest]

tests/tests_release.yml
ptoscano: fixed locally, will push once the CI completes (to check it works fine)

richm: ok - yeah, we won't need to do CI again after changing the comment

ptoscano:
Yay, it worked \o/

richm (approved):

PR #79, opened 2023-03-16
add and remove selinux ports for proxy access

richm:
The test proxy uses non-standard ports which are not in the default
SELinux policy, so add them for the test, and remove them at the
end of the test.

### DIFF
--- a/tests/tests_proxy.yml
+++ b/tests/tests_proxy.yml
@@ -6,12 +6,31 @@
   become: true
   tags:
     - tests::slow
+  vars:
+    __proxy_port_list:
+      - "{{ lsr_rhc_test_data.proxy_noauth_port }}"
+      - "{{ lsr_rhc_test_data.proxy_auth_port }}"
+      - "{{ lsr_rhc_test_data.proxy_nonworking_port }}"
   tasks:
     - name: Setup Candlepin
       import_tasks: tasks/setup_candlepin.yml
 
     - name: Setup Squid
       import_tasks: tasks/setup_squid.yml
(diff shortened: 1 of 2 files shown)


tests/tests_proxy.yml
```diff
@@ -314,3 +326,15 @@
         name: linux-system-roles.rhc
       vars:
         rhc_state: absent
+
+    - name: Remove SELinux policy for proxy ports
+      command: semanage port -d -t netport_port_t -p tcp {{ item }}
```
ptoscano, line 331:
```suggestion
      command: semanage port -d -t squid_port_t -p tcp {{ item }}
```

ptoscano (changes requested):
Thanks! Only the context needs to be changed. Also, maybe a bit more description in the commit message on the reason would be nice, so we don't forget.

ptoscano:
[citest]

ptoscano (approved):
LGTM - I tested the previous version with my proposed changes in, which is the current version.

I'll wait for the successful completion of the CI to merge.

ptoscano (changes requested):
@richm Hmm `policycoreutils-python-utils` needs to be installed for using `semanage` - can you please add the task for it in `tests_proxy`?

richm:
> @richm Hmm `policycoreutils-python-utils` needs to be installed for using `semanage` - can you please add the task for it in `tests_proxy`?

Instead, I'll just use the selinux role

ptoscano (approved):
OK!

PR #80, opened 2023-03-16
tests: refactor Insights setup

ptoscano:
Currently, the Insights-related tests assume they can work (as in, there is Insights support) when using an external test data configuration; this may not be true, e.g. when configuring to use a self-deployed Candlepin. Also, they all duplicate the logic for importing the test data.

Improve the situation a bit:
- add a `insights` boolean value for the test data to determine whether there is support for Insights; set it to false for the own Candlepin deployment
- create an helper `setup_insights.yml` to contain all the needed logic for setting up a test that uses Insights:
  - only the test data is set up, as an own Candlepin does not support Insights
  - end the play execution if the test data says there is no Insights support
  - check that Candlepin works (it's needed in any case)
- use the new `setup_insights.yml` in all the Insights-related tests.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -37,6 +37,7 @@ lsr_rhc_test_data:
   envs_register:
     - "Environment 1"
   env_nonworking: "not-a-valid-environment"
+  insights: false
 ```
 
 - `candlepin_host` & `candlepin_port` are the hostname & the port of Candlepin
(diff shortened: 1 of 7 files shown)


ptoscano:
[citest]

richm (approved):

PR #82, opened 2023-03-24
tests: updated insights unregister test to use rhc_insights instead r…

Archana-PandeyM:
…hc_state role variable

It appears that current insights unregister test is not validating rhc_insights variable, I have modified the test so that it can disconnect insights using-
rhc_insights:
  state: absent

### DIFF
--- a/tests/tests_insights_client_register.yml
+++ b/tests/tests_insights_client_register.yml
@@ -87,9 +87,17 @@
       include_role:
         name: linux-system-roles.rhc
       vars:
-        rhc_state: absent
+        rhc_insights:
+          state: absent
 
     - name: Unregister insights (noop)
+      include_role:
+        name: linux-system-roles.rhc
+      vars:
+        rhc_insights:
+          state: absent
+
+    - name: Unregister System
       include_role:
         name: linux-system-roles.rhc
       vars:


ahitacat (approved):
LGTM

richm:
[citest]

richm:
Looks like a lot of tests are failing in candlepin installation.

richm:
> Looks like a lot of tests are failing in candlepin installation.

It's probably not related to this change, but someone on the team should investigate.

Archana-PandeyM:
> > Looks like a lot of tests are failing in candlepin installation.
> 
> It's probably not related to this change, but someone on the team should investigate.

Archana-PandeyM:
@richm looks like canlepin failed to compile due to java version mismatch-
* What went wrong:
Execution failed for task ':compileJava'.
> error: invalid source release: 17
probably we need to fix java version
Candlepin deployment failure can be tracked as different issue

ptoscano:
Candlepin recently switched to OpenJDK 17, and the way we deploy it did not keep up with it. I'll have to look into it, it won't be a 5 minutes job...

In any case: the changes here affect only an Insights-related test, which is not run anyway using a self-deployed Candlepin; hence, this needs manual testing.

richm:
> Candlepin recently switched to OpenJDK 17, and the way we deploy it did not keep up with it. I'll have to look into it, it won't be a 5 minutes job...
> 
> In any case: the changes here affect only an Insights-related test, which is not run anyway using a self-deployed Candlepin; hence, this needs manual testing.

Then do we want to just skip those tests in upstream testing?

Archana-PandeyM:
> Candlepin recently switched to OpenJDK 17, and the way we deploy it did not keep up with it. I'll have to look into it, it won't be a 5 minutes job...
> 
> In any case: the changes here affect only an Insights-related test, which is not run anyway using a self-deployed Candlepin; hence, this needs manual testing.

I have manually tested on my local system.

ptoscano:
> > Candlepin recently switched to OpenJDK 17, and the way we deploy it did not keep up with it. I'll have to look into it, it won't be a 5 minutes job...
> > In any case: the changes here affect only an Insights-related test, which is not run anyway using a self-deployed Candlepin; hence, this needs manual testing.
> 
> Then do we want to just skip those tests in upstream testing?

They are already:
- `setup_insights.yml` sets up the test data only (no Candlepin)
  - the local test data has `insights: false`
- `setup_insights.yml` ends the execution if insights=false

richm (approved):

richm:
> > > Candlepin recently switched to OpenJDK 17, and the way we deploy it did not keep up with it. I'll have to look into it, it won't be a 5 minutes job...
> > > In any case: the changes here affect only an Insights-related test, which is not run anyway using a self-deployed Candlepin; hence, this needs manual testing.
> > 
> > 
> > Then do we want to just skip those tests in upstream testing?
> 
> They are already:
> 
>     * `setup_insights.yml` sets up the test data only (no Candlepin)
>       
>       * the local test data has `insights: false`
> 
>     * `setup_insights.yml` ends the execution if insights=false

Ok - I have approved and merged this PR - what can we do about the test failures?

PR #84, opened 2023-03-30
tests: switch to my unofficial Candlepin container image

ptoscano:
The current way to test registration is to deploy Candlepin from sources in the managed node; while this initially worked fine, it started to grow a number of drawbacks:
- building Candlepin from sources (including importing the test data for it) takes a lot of time, at least 5-6 minutes even on fast systems
- the Candlepin project changes the supported OSes over time, and thus it is not possible to test all the supported OSes of the role
- the currently used Ansible role is no more used/supported by the Candlepin project, and thus it is rotting In the end, we really don't need to build Candlepin from sources, but rather only use an already deployed Candlepin.

The Candlepin people are working on an official Candlepin container image usable also outside of their CI system; since this process will take some time, I built an unofficial container image of Candlepin:
  https://github.com/ptoscano/candlepin-container-unofficial
It contains everything (DB + Candlepin + test data) in a single image. While this is unofficial, it is not worse than the previously used way of deploying Candlepin, and it has the advantage of being stable (i.e. existing tags should still work fine regardless of upstream changes).

Hence, switch to this image for now:
- install podman to manage the container
- use the containers.podman to help interacting with podman
- start the container, waiting for Candlepin to start
- copy the certificates from the running container
- remove everything else related to the Candlepin deployment Also, since the OS needed for running Candlepin must not be the same OS of the managed node, the OS limitation is dropped.


ptoscano:
[citest]

ptoscano:
A bit of update/sum-up of the situation.

The approach generally works; the image I maintain is ~1.3 GB, and still downloading/running it is generally way much faster than deploying Candlepin.

There are few issues that need to be solved before landing this first:
- `Failed to register with 'localhost': com.redhat.RHSM1.Error: {"exception": "ValidationError", "severity": "error", "message": "Unknown arguments: dict_keys(['environment'])"}`
  - this is due to the renaming of the D-Bus option that happened when implementing multi-environments in subscription-manager, which landed in RHEL 8.6 and 9+
  - the new key is `environments`, and the D-Bus `Register*()` methods error out on unsupported options (sigh, yes, I'd have preferred not, but here we are...)
  - this affects CentOS Stream 8

- `Failed to register with 'localhost': com.redhat.RHSM1.Error: {"exception": "ConnectionOSErrorException", "severity": "error", "message": "('localhost', 8443, '/candlepin', OSError('Tunnel connection failed: 503 Service Unavailable',))"}`
  - the locally bound port 8443 to the 8443 port in the container seems to be IPv4-only by the podman default networking
  - the locally run squid proxy server listens to IPv6-only by default
  - since changing the networking of podman seems too much complicated, the likely solution here is to make squid listen to both IPv4 and IPv6
  - this affects all the distros

- Python traceback
  - this is due to D-Bus changing the behaviour of `unix:tmpdir=` >= 1.12.24 and >= 1.14.4: instead of abstract sockets, it will return local sockets
  - the D-Bus server in subscription-manager that listened to the returned address assumed it was an abstract socket, and thus it fails
  - subscription-manager >= 1.29.32 is fixed
  - this affects Fedora >= 35
  - the latest version of subscription-manager in Fedora is 1.29.30; I started the update process:
    - F39: should be available there soon (https://bodhi.fedoraproject.org/updates/FEDORA-2023-7e6b15703b)
    - F38: should be available in a week, and when the beta freeze is over (https://bodhi.fedoraproject.org/updates/FEDORA-2023-4ff782bf9c)
    - F37: should be available in a week (https://bodhi.fedoraproject.org/updates/FEDORA-2023-fff381197f)
    - F36: should be available in a week (https://bodhi.fedoraproject.org/updates/FEDORA-2023-8aa252fd1a)

- `Failed to register with 'localhost': com.redhat.RHSM1.Error: {"exception": "RestlibException", "severity": "error", "message": "Environment with ID \"Environment 2\" could not be found."}`
  - it seems like the newly created environment is not found, not sure why (needs more investigation)
  - this affects RHEL 8 (currently 8.9-nightly)

I _think_ that should cover all the issues I noticed so far; in case there are more, I want to weed out the existing ones first.

ptoscano:
>     * `Failed to register with 'localhost': com.redhat.RHSM1.Error: {"exception": "ConnectionOSErrorException", "severity": "error", "message": "('localhost', 8443, '/candlepin', OSError('Tunnel connection failed: 503 Service Unavailable',))"}`
>       
>       * the locally bound port 8443 to the 8443 port in the container seems to be IPv4-only by the podman default networking
>       * the locally run squid proxy server listens to IPv6-only by default
>       * since changing the networking of podman seems too much complicated, the likely solution here is to make squid listen to both IPv4 and IPv6
>       * this affects all the distros

Well, my attempts with squid did not lead to actual results...

That said, there was not much documentation on making podman listen to IPv6 interfaces too; the most useful resources I found so far are:
- https://github.com/containers/podman/issues/7415
- https://developers.redhat.com/articles/2022/08/10/how-conifgure-podman-40-ipv6

Since I'm not sure how messing up with the networks with different versions of podman in different versions of distros will work out, I'm going for the path of least resistance, i.e. use the "host" network for the container: it is slightly insecure (it will not use a separate network namespace), however I don't think it should be a problem in a throwaway test environment.

ptoscano:
[citest]

ptoscano:
[citest bad]

ptoscano:
Considering that most of the issues were due to
- old subscription-manager in Fedora that did not work well with recent dbus
- an invalid configuration with newer versions of squid (in Fedora)
- few issues in the `redhat_subscription` module (now temporarily worked around by pinning down the community.general collection)

Hence, I think all the issues concerning this switch should be solved now; hence, undrafting.

ptoscano:
[citest]

tests/tasks/setup_candlepin.yml
```diff
@@ -6,56 +6,32 @@
...
-            cp_configure_ruby: false
-            cp_deploy_args: "-g -a -f -t -r"
-            cp_git_ref: 9263a30b1612a22ec00ae05bd7d9ed6e0cdedbaf
-
-    - name: Enable tomcat
-      service:
-        name: tomcat
-        enabled: true
+    - name: Start Candlepin container
+      containers.podman.podman_container:
```
richm, line 26: Are we going to use this for downstream testing?  If so, I would rather not have another test dependency on an external, community supported collection.

ptoscano: it's used only when not using an external server (see `LSR_RHC_TEST_DATA` in `tests/README.md`); considering I manually call `podman` anyway, I guess I can avoid the usage of the collection and run `podman run` manually

ptoscano:
[citest]

ptoscano:
[citest]

ptoscano:
[citest]

richm (approved):

PR #86, opened 2023-04-03
Temporarily pin down community.general collection

ptoscano:
There are a couple of bugs in the `redhat_subscription` module shipped with the community.general 6.5.0 collection, plus it hits old versions of subscription-manager in Fedora.

As a short-gap measure to keep things working, pin down the version of community.general to anything lower than 6.5.0.


ptoscano:
[citest]

richm (approved):

PR #88, opened 2023-04-07
Fix minor ansible-lint issues

ptoscano:
- mark tasks that run `insights-client` as `changed_when=true`: this is done because, when they run, they will change things in the system
- fix indentation of conditions in a couple of `when` blocks


ptoscano:
[citest]

richm (approved):

PR #89, opened 2023-04-11
tests: update squid setup

ptoscano:
Improve the configuration of squid, especially in the authenticated setup, to handle better also newer versions:
- reduce the shutdown lifetime to 5s (from the 30s default): this is the time squid will wait on shutdown for connections; since all the connections are local, we can afford waiting only few seconds
- add the Candlepin port right before the first `Safe_ports` line, rather than before the last (it should place it closer to the existing `SSL_ports`)
- keep `http_access deny all` when authenticated: this is needed to deny all the connections that do not match any `http_access` rule before this
- configure the authentication parameters at the beginning of the file, so they are usable also mid-file
- declare the `authenticated` acl right after the ports, together with other acl's
- allow the `authenticated` acl before the global `deny all` rule, so it is properly used

### DIFF
--- a/tests/tasks/setup_squid.yml
+++ b/tests/tasks/setup_squid.yml
@@ -39,8 +39,15 @@
         path: /etc/squid/squid.conf
         regexp: "^acl SSL_ports port {{ lsr_rhc_test_data.candlepin_port }}"
         insertbefore: "^acl Safe_ports "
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

richm (approved):

PR #92, opened 2023-04-13
Do not pass fake creds when activation keys are specified

### LINKED ISSUE #90: Activation key registration fails if system is already registered
I'm using the rhc system role to register a system using an activation key.  When I run the playbook the first time, it succeeds and successfully registers the system.  

If I run the playbook again, it fails the second time with an error:

TASK [redhat.rhel_system_roles.rhc : Call subscription-manager] **********************************************************************************************
fatal: [rhel8-server2.example.com]: FAILED! => {"changed": false, "msg": "parameters are mutually exclusive: activationkey|username"}

Example playbook I'm using:
```yaml
- name: Run system roles
  hosts: all
  become: true
  tasks: 
    - name: run rhc system role
      include_role:
        name: redhat.rhel_system_roles.rhc
      vars:
        rhc_state: present
        rhc_auth:
          activation_keys:
            keys: 
              - "<redacted>"
        rhc_organization: "<redacted>"

```

ptoscano:
In case rhc_auth.activation_keys is specified (i.e. activation keys as credentials are used), then do not pass the fake username/password credentials to redhat_subscription, no matter the registration status; the redhat_subscription module rejects more than one credential type at the same time. This fixes specifying rhc_auth.activation_keys on an already registered system.

Check this case by duplicating the existing task for registration using activation keys in tests_register_unregister.yml, which now should pass (as noop) rather than failing.

Fixes #90

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -38,11 +38,13 @@
       else rhc_state | d('present') }}"
     username: "{{ __rhc_fake_credential
       if (rhc_state | d('present') == 'present'
-           and __rhc_subman_identity.rc == 0)
+           and __rhc_subman_identity.rc == 0
+           and 'activation_keys' not in rhc_auth)
       else (rhc_auth.login.username | d(omit)) }}"
     password: "{{ __rhc_fake_credential
       if (rhc_state | d('present') == 'present'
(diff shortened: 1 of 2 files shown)


ptoscano:
[citest]

PR #94, opened 2023-04-17
Minimize Ansible facts needed

ptoscano:
The role requires only a couple of Ansible facts, and it has already some logic to handle their lack; sadly, since all the tests have `gather_facts=true`, this is not tested (as all the Ansible facts are available upfront). Also, the current way for the role to ensure to have its own facts is to collect the "min" subset, which contains way more than what the role needs.

Improve this situation a bit, both in what the role needs and how the tests work:
- introduce a separate `__rhc_required_fact_subsets` internal variable, which represents the list of fact subsets matching the existing `__rhc_required_facts` internal variable; they look similar now, they don't necessarily need to, however
- use `__rhc_required_fact_subsets` for the "setup" task of the role, rather than "min"
- set gather_facts=false in all the tests to ensure no facts are collected upfront, to check that what is needed will be collected on demand before use
- since "tests_environments" explicitly uses a couple of Ansible facts, manually collect them
- collect also a couple of facts when using external test data: this is because it is possible to conditionalize some of the test data variables based on the distribution version

This should make the role slightly faster to run (at least the first time in a play) in case none of the needed Ansible facts for it is already collected.


tasks/set_vars.yml
```diff
@@ -2,7 +2,8 @@
 ---
 - name: Ensure ansible_facts used by role
   setup:
-    gather_subset: min
+    gather_subset:
```
richm, line 5:
```suggestion
    gather_subset: "{{ __rhc_required_fact_subsets }}"
```

tasks/set_vars.yml
```diff
@@ -2,7 +2,8 @@
...
+      "{{ __rhc_required_fact_subsets }}"
```
richm, line 6:
```suggestion
```

richm:
> @richm thanks for the notes, I have the fixes locally. Before I keep testing/updating this: do you see anything problematic in the approach/changes?

No, I think this is fine.

richm:
[citest]

richm (approved):

PR #95, opened 2023-04-24
ci: Add commitlint GitHub action to ensure conventional commits

spetrosi:
For more information, see Conventional Commits format in Contribute
https://linux-system-roles.github.io/contribute.html#conventional-commits-format

Signed-off-by: Sergei Petrosian <spetrosi@redhat.com>

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,44 @@
+name: Commitlint
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - edited
+  merge_group:
+    branches:
+      - main
+    types:
+      - checks_requested
+permissions:
+  contents: read
+jobs:
+  commit-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install conventional-commit linter
+        run: npm install @commitlint/config-conventional @commitlint/cli
+
+      # Finding the commit range is not as trivial as it may seem.
+      #
+      # At this stage, git's HEAD does not refer to the latest commit in the
+      # PR, but rather to the merge commit inserted by the PR. So instead we
+      # have to get 'HEAD' from the PR event.
+      #
+      # One cannot use the number of commits
+      # (github.event.pull_request.commits) to find the start commit
+      # i.e. HEAD~N does not work, this breaks if there are merge commits.
+      - name: Run commitlint on commits
+        run: >-
+          npx commitlint --from '${{ github.event.pull_request.base.sha }}'
+          --to '${{ github.event.pull_request.head.sha }}' --verbose
+
+      - name: Run commitlint on PR title
+        run: >-
+          echo '${{ github.event.pull_request.title }}' |
+          npx commitlint --verbose
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,0 +1,141 @@
+module.exports = {
+	parserPreset: 'conventional-changelog-conventionalcommits',
+	rules: {
+		'body-leading-blank': [1, 'always'],
+		'body-max-line-length': [2, 'always', 100],
+		'footer-leading-blank': [1, 'always'],
+		'footer-max-line-length': [2, 'always', 100],
+		'header-max-length': [2, 'always', 100],
+		'subject-case': [
+			2,
+			'never',
+			['start-case', 'pascal-case', 'upper-case'],
+		],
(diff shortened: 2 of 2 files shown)


.ansible-lint
```diff
@@ -18,5 +18,6 @@ exclude_paths:
   - tests/roles/
   - .github/
   - examples/roles/
+  - tests/roles/candlepin/
```
ptoscano, line 21: this was removed, and it does not need to be brought back

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,60 @@
...
+          echo "module.exports = {
+            extends: ['@commitlint/config-conventional'],
+            rules: {
+              'subject-case': [
+                2,
+                'never',
+                ['start-case', 'pascal-case', 'upper-case'],
+              ],
+            }
+          }" > commitlint.config.js
```
ptoscano, line 38: this makes this config buried in the action, recreating it every time, and also requiring escaping (as it's inside a shell command); please rather save it wouldn't it be better to have it as e.g. `.commitlintrc.js` file in the repository? (according to [1] it seems supported) this way it's easier to review/edit, and it's hidden by default (like `.ansible-lint`, `.yamllint.yml`, etc)

[1] https://github.com/conventional-changelog/commitlint#config

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,60 @@
...
+        #
+        # One cannot use the number of commits
+        # (github.event.pull_request.commits) to find the start commit
+        # i.e. HEAD~N does not work, this breaks if there are merge commits.
+        run: |
+          echo "PR_HEAD=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+          echo "PR_BASE=${{ github.event.pull_request.base.sha }}" >> $GITHUB_ENV
+
+      - name: Run commitlint on commits
+        run: npx commitlint --from $PR_BASE --to $PR_HEAD --verbose
```
ptoscano, line 55: rather than setting envvars from existing vars, simple use what is there in the command? ie
```yaml
      - name: Run commitlint on commits
        run: npx commitlint --from "${{ github.event.pull_request.base.sha }}" --to "${{ github.event.pull_request.head.sha }}" --verbose
```

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Run commitlint on PR title
+        run: |
+          echo ${{ github.event.pull_request.title }} | npx commitlint --verbose
```
ptoscano, line 59: shouldn't the PR title be quoted, since this is a shell command?

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,60 @@
...
+
```
ptoscano, line 60: nitpick: extra empty line at the end of the file

ptoscano (changes requested):

.github/workflows/commitlint.yml
richm: My thinking is that since the config is small, it isn't a big deal to just create it in the github action, and save having to create another file in https://github.com/linux-system-roles/.github/tree/main/playbooks/templates and manage it in https://github.com/linux-system-roles/.github/blob/main/inventory/group_vars/active_roles.yml

The reason we have `.ansible-lint` and `.yamllint.yml` are historical - although having an `.ansible-lint` is probably a good thing since it is larger and more complex, but `.yamllint.yml` is almost trivially small.

.github/workflows/commitlint.yml
richm: it should be quoted in case `github.event.pull_request.title` contains a `*` or other shell metacharacter.
```suggestion
          echo '${{ github.event.pull_request.title }}' | npx commitlint --verbose
```

ptoscano:
Thanks for the updates -- having a 130+ lines static `.commitlintrc.js` is indeed much better than burying it in the action steps!

I'm personally not fond of the Conventional Commits style; that said, since it's part of the Linux System Roles contribution guidelines, I'll adapt to it. I have only one small request: for test-only changes I've been using "tests: " as prefix so far; would it be possible to accept it in addition to "test: "?

spetrosi:
> Thanks for the updates -- having a 130+ lines static `.commitlintrc.js` is indeed much better than burying it in the action steps!
Agreed!
> 
> I'm personally not fond of the Conventional Commits style; that said, since it's part of the Linux System Roles contribution guidelines, I'll adapt to it. I have only one small request: for test-only changes I've been using "tests: " as prefix so far; would it be possible to accept it in addition to "test: "?
Done that, `tests:` can be used as a commit type now as well as `test:`

PR #102, opened 2023-05-23
fix: fix filename with insights-client tags

ptoscano:
The file with the tags of insights-client is
`/etc/insights-client/tags.yaml`, not `tags.yml`.

To ease the situation a bit, move the full path to a role variable.

Reported-by: Benjamin Blasco <bblasco@redhat.com>

### DIFF
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -12,7 +12,7 @@
 
 - name: Add insights tags
   copy:
-    dest: /etc/insights-client/tags.yml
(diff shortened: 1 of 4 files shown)


ptoscano:
Since this test cannot run in CI (requires a real Insights server), I tested it manually. The result is that tags are properly shown for the hosts :heavy_check_mark:

ptoscano:
[citest]

richm (approved):

PR #104, opened 2023-05-31
build: require community.general 6.2.0

ptoscano:
This version contains a number of features and fixes that greatly help this role; while the role was able so far to work with older versions, let's require a recent version so it is possible to drop workarounds.

The upper version limit is unfortunately still needed for now.


ptoscano:
[citest]

PR #105, opened 2023-06-01
refactor: pass rhc_baseurl directly to redhat_subscription

ptoscano:
The `redhat_subscription` used to handle certain options both writing them to the config file (using `subscription-manager config`), and passing them as command line parameters to `subscription-manager register`. This was problematic for `rhsm_baseurl`, as subscription-manager sadly forces `https` as scheme when parsing `--baseurl` as passed to `register`. This was fixed with [1] (backported to community.general 6.x with [2]), available since community.general 6.1.0.

Hence, directly pass the `rhc_baseurl` role parameter as `rhsm_baseurl` parameter of the `redhat_subscription` module.

[1] https://github.com/ansible-collections/community.general/pull/5583
[2] https://github.com/ansible-collections/community.general/pull/5626

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -17,19 +17,6 @@
   changed_when: false
   failed_when: __rhc_subman_identity.rc not in [0, 1]
 
-- name: Change rhsm.baseurl
-  command:
-    argv:
-      - subscription-manager
-      - config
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

richm (approved):

PR #108, opened 2023-06-02
fix: correctly set release using new behavior of rhsm_release

richm:
The way the new version of rhsm_release works is - if the `release`
parameter is present, set the release to the given value.
If there is no `release` parameter, unset the release.
If the user does not set `rhc_release`, then it will be `null`,
and calling the rhsm_release module will be skipped.
See https://github.com/ansible-collections/community.general/pull/6401
for details.

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -46,7 +46,9 @@
           and not 'keys' in rhc_auth.activation_keys | d({})) else omit }}"
     release: "{{ rhc_release
       if (rhc_state | d('present') != 'absent'
-           and rhc_release != __rhc_state_absent)
+           and rhc_release != __rhc_state_absent
+           and not rhc_release is none
+           and rhc_release != omit)
       else omit }}"
     server_proxy_hostname: "{% if rhc_proxy == __rhc_state_absent -%}
       {{ __rhc_empty_string }}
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

ptoscano:
It'd seem OK at a quick glance; i need to test this manually using `tests_release.yml`, as a self-deployed Candlepin has no concept of "releases".

ptoscano:
Unfortunately this does not fix the situation completely, and unsetting a release (i.e. specifying `rhc_release {"state":"absent"}`) is still broken when using ansible-core 2.15:
```
argument 'release' is of type <class 'NoneType'> and we were unable to convert to str: 'None' is not a string and conversion is not allowed
```

The situation is the following:
- ansible-core 2.15 rejected `null` as valid parameter: https://github.com/ansible/ansible/pull/79677
- ... so `rhsm_release` was changed in `community.general` 6.6.0 to not require the `release` parameter, unsetting the release if not specified: https://github.com/ansible-collections/community.general/pull/6401
- ... so we would adapt to that (and bump the requirement to `community.general >= 6.6.0`, however our version is pinned to `< 6.5.0` for other reasons: #86
- ... so we would unpin the requirement, however it is not possible yet as it would break Fedora managed nodes: #91
- fixing Fedora means fixing the SELinux policy: https://github.com/fedora-selinux/selinux-policy/pull/1726
- the above PR currently hits the temporary `dnf.conf` screwup in Rawhide: https://bugzilla.redhat.com/show_bug.cgi?id=2210694

Yes, a giant mess.

ptoscano:
> Yes, a giant mess.

Let's untangle it by temporarily dropping the support for Fedora: #109.

ptoscano:
OK, created #110 so we can rely on a newer `rhsm_release` module from community.general.

Because of the new behaviour, we can simplify things a bit: instead of two `rhsm_release` invocations (depending on whether we need to set or unset a release), they can be merged into one, i.e. passing the "release" parameter when setting or `omit`ing it when unsetting.

richm:
[citest]

tasks/subscription-manager.yml
```diff
@@ -93,21 +94,14 @@
...
-        - rhc_release != __rhc_state_absent
-
-    - name: Unset the release
-      community.general.rhsm_release:
-        release: null
-      when:
-        - rhc_state | d("present") == "present"
-        - __rhc_subman_identity.rc == 0
-        - rhc_release == __rhc_state_absent
+        - not rhc_release is none
```
ptoscano, line 104: in addition to this, most likely you need to check that `rhc_release != omit`; i remember it was needed when the hack for `rhc_baseurl` was needed (see #12); even more so, because
```yaml
  foo: null
...
  rhc_release: "{{ foo | d(omit) }}"
```
will then result in the special `omit-RANDOMSTRING` passed to the role, which means it will be passed to "release:" of `rhsm_release`, and thus unset the release rather than ignoring it - or am i wrong?

PR #111, opened 2023-06-06
refactor: drop fake credential workaround

ptoscano:
The `redhat_subscription` module used to always require credentials when `state=present`; since we used the module also to do basic configuration of `subscription-manager`, this required us to pass fake credentials (unused) when the system was already registered. This behaviour was changed with [1] (backported to `community.general` 6.x with [2]), and available since `community.general` 6.5.0.

Hence, drop the concept of fake credential, and pass `username` and `password` only when specified; `redhat_subscription` still properly checks that only one credential type is specified.

[1] https://github.com/ansible-collections/community.general/pull/5664
[2] https://github.com/ansible-collections/community.general/pull/6222

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -23,16 +23,8 @@
     state: "{{ 'present'
       if (rhc_state | d('present') == 'reconnect')
       else rhc_state | d('present') }}"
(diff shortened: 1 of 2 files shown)


ptoscano:
[citest]

richm (approved):

PR #116, opened 2023-06-07
fix: enable remediation only on RHEL >= 8.4

ptoscano:
The rhc-worker-playbook package, needed for the remediation support, is available in RHEL only starting from 8.4 (including 9); hence, do not attempt to setup remediation on older RHEL versions.

### DIFF
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -58,7 +58,9 @@
   changed_when: true
 
 - name: Configure remediation
-  when: rhc_insights.remediation | d("present") == "present"
+  when:
+    - ansible_distribution_version is version("8.4", ">=")
+    - rhc_insights.remediation | d("present") == "present"
   block:
     - name: Ensure worker and ansible packages are installed
       package:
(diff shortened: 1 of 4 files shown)


tasks/insights-client.yml
```diff
@@ -58,7 +58,9 @@
   changed_when: true
 
 - name: Configure remediation
-  when: rhc_insights.remediation | d("present") == "present"
+  when:
+    - ansible_distribution_version is version('8.4', '>=')
```
richm, line 62:
```suggestion
    - ansible_distribution_version is version("8.4", ">=")
```
prefer double quotes, or at least be consistent with other quoting in surrounding code

tests/tests_insights_remediation.yml
```diff
@@ -8,6 +8,18 @@
...
+      setup:
+        gather_subset:
+          - "!all"
+          - "!min"
+          - distribution_version
+
+    - name: Skip if the RHEL version is older than 8.4
+      meta: end_play
+      when:
+        - ansible_distribution_version is version('8.4', '<')
```
richm, line 21:
```suggestion
        - ansible_distribution_version is version("8.4", "<")
```

richm (approved):

PR #117, opened 2023-06-08
tests: tests_environments: fix distro version check

ptoscano:
Since we know which version can list the environments, specify that.

Fixes this test when run with RHEL 8 < 8.6.


ptoscano:
[citest]

richm (approved):

PR #120, opened 2023-06-09
tests: disable rhc_insights.autoupdate when not needed

ptoscano:
Disable the Insights autoupdate in all the tests that do not deal with autoupdate explicitly.

The two tests run fine when run with Stage.


ptoscano:
We actually want `autoupdate` on by default when testing Insights, as otherwise `insights-client` will not download the newest egg from the Insights server.

Thanks @ahitacat for the (offline) clarification!

PR #121, opened 2023-06-09
ci: Add pull request template and run commitlint on PR title only

spetrosi:
We now ensure the conventional commits format only on PR titles and not on
commits to let developers keep commit messages targeted for other developers
i.e. describe actual changes to code that users should not care about.
And PR titles, on the contrary, must be aimed at end users.

For more info, see
https://linux-system-roles.github.io/contribute.html#write-a-good-pr-title-and-description

Signed-off-by: Sergei Petrosian <spetrosi@redhat.com>

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+Enhancement:
+
+Reason:
+
+Result:
+
+Issue Tracker Tickets (Jira or BZ if any):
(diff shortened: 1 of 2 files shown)


ptoscano:
>  And PR titles, on the contrary, must be aimed at end users.

Hmm? I don't understand this, can you please describe it?

spetrosi:
I am sorry for a lot of noise on this, turned out that how I planned it initially does not work so good, so I am applying more changes. I'll send an email on this soon.
The drawback of the previous approach is that each commit must go to the changelog, but developers write commit messages for other developers, not for end users. You want the commit to state clear why the code was change, and often commit messages include information that end users do not care about. So it makes more sense to keep commit messages for developers, and move the user-aimed information in PR description and title.

PR #124, opened 2023-07-13
fix: facts being gathered unnecessarily

richm:
Cause: The comparison of the present facts with the required facts is
being done on unsorted lists.

Consequence: The comparison may fail if the only difference is the
order.  Facts are gathered unnecessarily.

Fix: Use `difference` which works no matter what the order is.  Ensure
that the fact gathering subsets used are the absolute minimum required.

Result: The role gathers only the facts it requires, and does
not unnecessarily gather facts.

### DIFF
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -12,12 +12,8 @@ __rhc_required_facts:
 # the subsets of ansible_facts that need to be gathered in case any of the
 # fact in __rhc_required_facts is missing; see the documentation of
 # the 'gather_subset' parameter of the 'setup' module
-__rhc_required_fact_subsets:
-  - "!all"
-  - "!min"
-  - distribution
-  - distribution_major_version
-  - distribution_version
+__rhc_required_fact_subsets: "{{ ['!all', '!min'] +
+  __rhc_required_facts }}"
 
 __rhc_state_absent:
   state: absent
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -3,8 +3,8 @@
 - name: Ensure ansible_facts used by role
   setup:
     gather_subset: "{{ __rhc_required_fact_subsets }}"
-  when: not ansible_facts.keys() | list |
-    intersect(__rhc_required_facts) == __rhc_required_facts
+  when: __rhc_required_facts |
+    difference(ansible_facts.keys() | list) | length > 0
 
 - name: Check if insights-packages are installed
   package_facts:


richm:
[citest]

vars/main.yml
```diff
@@ -12,12 +12,8 @@ __rhc_required_facts:
...
 # fact in __rhc_required_facts is missing; see the documentation of
 # the 'gather_subset' parameter of the 'setup' module
-__rhc_required_fact_subsets:
-  - "!all"
-  - "!min"
-  - distribution
-  - distribution_major_version
-  - distribution_version
+__rhc_required_fact_subsets: "{{ ['!all', '!min'] +
+  __rhc_required_facts }}"
```
ptoscano, line 16: this works right now because of the content of `__rhc_required_facts`, however it is not guaranteed to be true in the future; this is a list of categories used by `setup.gather_subset`, whereas `__rhc_required_facts` is a list of actual fact variables

as much as it seems like a duplication, i'd keep the two lists separate

ptoscano (changes requested):
the change to the `when` condition of `setup` is OK; note I copied it from the template, so maybe other roles might require the same fix

vars/main.yml
richm: I'd like to use the method I used in https://github.com/linux-system-roles/postgresql/pull/43:
```yaml
# these facts cannot be used as a gather_subset
__postgresql_no_subset_facts: [ansible_memory_mb]

# the subsets of ansible_facts that need to be gathered in case any of the
# facts in required_facts is missing; see the documentation of
# the 'gather_subset' parameter of the 'setup' module
__postgresql_required_facts_subsets: "{{ ['!all', '!min', 'hardware'] +
  __postgresql_required_facts | difference(__postgresql_no_subset_facts) }}"
```
In every other case in every other system role, the list of `__xxx_required_facts` could be used in `__xxx_required_facts_subset` - I don't want to be in a situation where we have to maintain the same content in both `__xxx_required_facts` and `__xxx_required_facts_subset` - I think this solution will be the easiest to maintain going forward.

ptoscano: this works only when using fact variables that are the same as the facts "categories" of the `setup` module, though; in the case of `rhc`, the facts used are very few, however i wanted to avoid implicit assumptions like that...

richm: > this works only when using fact variables that are the same as the facts "categories" of the `setup` module, 

Right - but in practice, in every case I have found in every system role except postgresql, the name of the fact variable is the same as the name of a fact "category".

> though; in the case of `rhc`, the facts used are very few, however i wanted to avoid implicit assumptions like that...

Sure - but I think the tradeoff here is - if we have to maintain two separate lists of (practically) identical content, we may miss something if we need to add a dependency on a new fact - but if we use one list for both, and we add a required fact that isn't a category, then the `setup:` module call will exit with a descriptive error, so it will be caught immediately, and the remedy is straightforward.

PR #144, opened 2023-10-17
tests: use the containers.podman collection

ptoscano:
Make use of the `containers.podman` collection to ease the management of containers: in particular, the `podman_container` module is used to start the container of Candlepin. While the result is the same, using the `podman_container` module eases the handling of the parameters for the containers, which are now typed variables rather than untyped command line arguments.

There is no behaviour change.


ptoscano:
[citest]

PR #146, opened 2023-11-22
tests: start candlepin container as privileged on EL 7

ptoscano:
It seems to be needed in EL 7; it should not be a problem, since the tests are usually run on an isolated system.

Ensure to have the right distro facts for the needed check.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -6,6 +6,14 @@
 - name: Deploy Candlepin
   when: lsr_rhc_test_data_file | length == 0
   block:
+    - name: Get facts for external test data
+      setup:
+        gather_subset:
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

ukulekek:
[citest]

richm:
ok - but note that ci tests will not run on el7 because "7" is not listed as a supported platform at https://github.com/linux-system-roles/rhc/blob/main/meta/main.yml#L13

ptoscano:
> note that ci tests will not run on el7 because "7" is not listed as a supported platform

Yes, that is fine - the actual enablement will come after this.

I sent this because it was already a known bit I needed to change for it; in the meanwhile I tested whether there was anything else needed, and so far it seems not.

PR #147, opened 2023-11-29
refactor: improve support for ostree systems

richm:
The dependency on `ansible.utils.update_fact` is causing issue with
some users who now must install that collection in order to run
the role, even if they do not care about ostree.

The fix is to stop trying to set `ansible_facts.pkg_mgr`, and instead
force the use of the ostree package manager with the `package:` module
`use:` option.  The strategy is - on ostree systems, set the flag
`__$ROLENAME_is_ostree` if the system is an ostree system.  The flag
will either be undefined or `false` on non-ostree systems.
Then, change every invocation of the `package:` module like this:

```yaml
- name: Ensure required packages are present
  package:
    name: "{{ __$ROLENAME_packages }}"
    state: present
    use: "{{ (__$ROLENAME_is_ostree | d(false)) |
      ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
```

This should ensure that the `use:` parameter is not used if the system
is non-ostree.  The goal is to make the ostree support as unobtrusive
as possible for non-ostree systems.
The user can also set `__$ROLENAME_is_ostree: true` in the inventory or play
if the user knows that ostree is being used and wants to skip the check.
Or, the user is concerned about the performance hit for ostree detection
on non-ostree systems, and sets `__$ROLENAME_is_ostree: false` to skip
the check.
The flag `__$ROLENAME_is_ostree` can also be used in the role or tests to
include or exclude tasks from being run on ostree systems.

This fix also improves error reporting in the `get_ostree_data.sh` script
when included roles cannot be found.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -22,5 +22,3 @@ exclude_paths:
   - examples/roles/
 mock_roles:
   - linux-system-roles.rhc
-mock_modules:
-  - ansible.utils.update_fact
--- a/.ostree/get_ostree_data.sh
+++ b/.ostree/get_ostree_data.sh
@@ -2,7 +2,6 @@
 
 set -euo pipefail
 
-role_collection_dir="${ROLE_COLLECTION_DIR:-fedora/linux_system_roles}"
 ostree_dir="${OSTREE_DIR:-"$(dirname "$(realpath "$0")")"}"
 
 if [ -z "${4:-}" ] || [ "${1:-}" = help ] || [ "${1:-}" = -h ]; then
@@ -29,7 +28,7 @@ if [ "$pkgtype" = testing ]; then
 fi
 
 get_rolepath() {
-    local ostree_dir role rolesdir roles_parent_dir
+    local ostree_dir role rolesdir roles_parent_dir coll_path path
     ostree_dir="$1"
     role="$2"
     roles_parent_dir="$(dirname "$(dirname "$ostree_dir")")"
@@ -47,16 +46,22 @@ get_rolepath() {
         fi
     done
     # look elsewhere
-    if [ -n "${ANSIBLE_COLLECTIONS_PATHS:-}" ]; then
-        for pth in ${ANSIBLE_COLLECTIONS_PATHS//:/ }; do
-            rolesdir="$pth/ansible_collections/$role_collection_dir/roles/$role/.ostree"
-            if [ -d "$rolesdir" ]; then
-                echo "$rolesdir"
-                return 0
-            fi
+    coll_path="${ANSIBLE_COLLECTIONS_PATH:-}"
+    if [ -z "$coll_path" ]; then
+        coll_path="${ANSIBLE_COLLECTIONS_PATHS:-}"
+    fi
+    if [ -n "${coll_path}" ]; then
+        for path in ${coll_path//:/ }; do
+            for rolesdir in "$path"/ansible_collections/*/*_system_roles/roles/"$role"/.ostree; do
+                if [ -d "$rolesdir" ]; then
+                    echo "$rolesdir"
+                    return 0
+                fi
+            done
         done
     fi
-    return 1
+    1>&2 echo ERROR - could not find role "$role" - please use ANSIBLE_COLLECTIONS_PATH
+    exit 2
 }
 
 get_packages() {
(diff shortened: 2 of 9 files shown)


richm:
[citest]

ptoscano (commented):
```yaml
    use: "{{ (__rhc_is_ostree | d(false)) |
             ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
```
I see a couple of problems with this:
- from what I can see, `__rhc_is_ostree` seems to be always set, either via `set_vars.yml`, or in the tests setup; thus, both `d()` and `ternary()` seem unneeded?
- lot of copy&paste everywhere we need to install packages

What I'd do is: instead of a `__rhc_is_ostree` bool variable, create a `__rhc_pkgman` (or something) being either `omit` or `auto` by default, setting it to `ansible.posix.rhel_rpm_ostree` in case of ostree systems; this way, that variable is always defined, and it can be passed directly to `package`:
```yaml
    use: "{{__rhc_pkgman }}"
```
It's still something you need to manually apply when using `package` in a new place, however IMHO it is much more compact, less copy&paste, and nicer to read.

richm:
> ```yaml
>     use: "{{ (__rhc_is_ostree | d(false)) |
>              ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
> ```
> 
> I see a couple of problems with this:
> 
>     * from what I can see, `__rhc_is_ostree` seems to be always set, either via `set_vars.yml`, or in the tests setup; thus, both `d()` and `ternary()` seem unneeded?

ternary is needed - `d` is not

> 
>     * lot of copy&paste everywhere we need to install packages
> 
> 
> What I'd do is: instead of a `__rhc_is_ostree` bool variable, create a `__rhc_pkgman` (or something) being either `omit` or `auto` by default, setting it to `ansible.posix.rhel_rpm_ostree` in case of ostree systems; this way, that variable is always defined, and it can be passed directly to `package`:
> 
> ```yaml
>     use: "{{__rhc_pkgman }}"
> ```
> 
> It's still something you need to manually apply when using `package` in a new place, however IMHO it is much more compact, less copy&paste, and nicer to read.

ok - but 
* many roles still need a boolean `__$ROLENAME_is_ostree` for other reasons and I wanted to make the implementation consistent across all of the 28 roles
* I didn't want to define two different variables - one for `is_ostree` and one for `pkgman`
* it doesn't eliminate the copy/paste, and requirement to remember to add `use:` to every `package:` module from now and the future, it just makes it a little cleaner

richm:
I tried using

```yaml
    - name: Set package manager for ostree or omit
      set_fact:
        __rhc_pkg_mgr: "{{ __ostree_booted_stat.stat.exists |
          ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
```

I got the following error:

```
TASK [linux-system-roles.rhc : Set package manager for ostree or omit] *********
task path: /home/rmeggins/linux-system-roles/rhc/tests/roles/linux-system-roles.rhc/tasks/set_vars.yml:17
Friday 01 December 2023  16:25:14 -0700 (0:00:01.185)       0:00:04.553 ******* 
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NoneType: None
fatal: [/home/rmeggins/.cache/linux-system-roles/rhel-8-y.qcow2]: FAILED! => {
    "changed": false
} 
  
MSG:

No key/value pairs provided, at least one is required for this action to succeed
```

because the `omit` is processed by the module in which it is used, or at least by the `set_fact` module which is notorious for expanding and evaluating every variable and reference (as opposed to `vars` which are lazily evaluated)

However, this works:

```yaml
    - name: Set package manager for ostree or auto
      set_fact:
        __rhc_pkg_mgr: "{{ __ostree_booted_stat.stat.exists |
          ternary('ansible.posix.rhel_rpm_ostree', 'auto') }}"
```

because the default value for `use:` is `auto` - see https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html#parameter-use and https://docs.ansible.com/ansible/2.9/modules/package_module.html#parameter-use

This is slightly more invasive than using `use: .... , omit)` because if the default changes to something other than `auto` in the future, the code would need to do something like

```yaml
    - name: Set package manager for ostree or omit
      set_fact:
        __rhc_pkg_mgr: "{{ __ostree_booted_stat.stat.exists |
          ternary('ansible.posix.rhel_rpm_ostree', default_mgr) }}"
      vars:
        default_mgr: "{{ (ansible_version.full is version('2.18', '>=')) | ternary('not-auto', 'auto') }}"
```
which I suppose is not too terrible, but it does mean that the role could potentially break for users who upgrade to a newer version of ansible, and worse, it would break because of ostree for users who don't care about ostree.

richm:
@ptoscano any further objections, or can this be merged?

PR #151, opened 2023-12-07
feat: support again EL7

ptoscano:
Enhancement:
After yet another internal discussion, re-enable the support for EL7; no other changes are needed in the role.

https://issues.redhat.com/browse/RHEL-16976
https://issues.redhat.com/browse/RHEL-16977

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -12,6 +12,7 @@ galaxy_info:
   platforms:
     - name: EL
       versions:
+        - "7"
         - "8"
         - "9"
 


ptoscano:
[citest]

richm (approved):

ptoscano:
[citest bad]

ptoscano:
> NOTE that baseos ci will reuse the same VM for each test.

Hm is it really the case? Looking at the output of each test that actually runs (e.g. not the Insights-related tests), it seems to install `podman` again, which means the guest is clean... except for the last test, i.e. `tests_repositories.yml`.

Regarding the behaviour of `podman_container`: considering that it uses `state: started`, then I don't understand why it tries to create the container again, rather than using the existing running container. According to the doc:
> _started_ - Asserts there is a running container matching the name and any provided configuration. If no container matches the name, a container will be created and started. Use recreate to always re-create a matching container, even if it is running. Use force_restart to force a matching container to be stopped and restarted.

richm:
I guess the version of podman on el7 is just too old for the module :-(

This is how I get all of the tests to pass:
```
@@ -32,6 +32,11 @@
         use: "{{ __rhc_is_ostree |
                  ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
+    - name: Stop and remove Candlepin container
+      containers.podman.podman_container:
+        name: candlepin
+        state: absent
+
     - name: Start Candlepin container
       containers.podman.podman_container:
```
testing on centos-9 now

richm:
centos-9 passes - so it looks like the right way is to set the container `state: absent` first

ptoscano:
Let try then: #153.

richm:
> Let try then: #153.

merged

richm (approved):

PR #152, opened 2023-12-08
tests: force recreating Candlepin container

ptoscano:
It seems that the `podman_container` thinks that the container is changed, in case it is already running, and thus it tries to recreate it; this might be an already reported issue [1].

To avoid problems on this, force the creation of the container to a new one, in case it is running already. This also gives back a new clean Candlepin.

[1] https://github.com/containers/ansible-podman-collections/issues/532


ptoscano:
[citest]

PR #153, opened 2023-12-11
tests: manually stop the Candlepin container

ptoscano:
It seems the `podman_container` module cannot properly stop the running container when being asked using `recreate=true`; hence, do it manually before starting the new container.

This also reverts commit 12c1a38bd914547986509dbd1b9c955bc0f838f9, which is no more needed now.


ptoscano:
[citest]

richm (approved):

PR #155, opened 2023-12-12
feat: add ansible host parameter to insights configuration

DuckBoss:
Adds the ansible host name option for insights-client
into the rhc role. This adds  support for setting an
ansible host name in inventory during insights-client
registration and updating the current ansible host name
in inventory if the system is already registered.

**Changes:**
 - Adds `ansible_host` parameter to insights client task to set/update/reset ansible host name in HBI.
 - Adds tests for ansible host name configuration.
 - Updates README with usage documentation for `ansible_host` parameter.

Signed-off-by: Jason Jerome <jajerome@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,21 @@ enabled by default.
 
 ```yaml
     rhc_insights:
+      ansible_host: "example-host"
+```
+
+Configures the ansible host name with a custom value for the system record
+in Host Based Inventory (HBI). This host name is used in playbooks by remediations.
+It is `null` by default and will use the system host name if not specified.
+
+Possible values of this variable:
+
+* `null` or an empty value: the ansible host name is not changed.
+* `{state: absent}`: the ansible host name is unset in the insights-client config file and Host Based Inventory (HBI) is updated to use the system host name.
+* any other string value: the ansible host name is changed in Host Based Inventory (HBI).
+
+```yaml
+    rhc_insights:
       remediation: present
 ```
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -4,6 +4,7 @@ rhc_auth: {}
 rhc_baseurl: null
 rhc_environments: []
 rhc_insights:
+  ansible_host: null
   autoupdate: true
   remediation: present
   state: present
--- a/tasks/insights-client-unregistration.yml
+++ b/tasks/insights-client-unregistration.yml
@@ -11,3 +11,9 @@
       or "Registered" in __rhc_insights_status.stdout
   register: __rhc_insights_unregister_result
   changed_when: true
+
+- name: Reset ansible host in config
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^ansible_host="
+    state: absent
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -38,6 +38,62 @@
     insertafter: "#auto_update"
     line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
 
+- name: Check ansible host in insights-client config
+  when:
+    - rhc_insights.ansible_host is defined
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != __rhc_state_absent
+    - rhc_insights.ansible_host != omit
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^ansible_host="
+    state: present
+    line: ansible_host={{ rhc_insights.ansible_host }}
+  check_mode: true
+  register: __insights_ansible_host_exists
+
+- name: Configure ansible host
+  when:
+    - rhc_insights.ansible_host is defined
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != omit
+    - rhc_insights.ansible_host != __rhc_state_absent
+    - __insights_ansible_host_exists.changed
+  block:
+    - name: Update ansible host in insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^ansible_host="
+        insertafter: "#ansible_host="
+        line: ansible_host={{ rhc_insights.ansible_host }}
+    - name: Update ansible host in inventory
+      shell: >-
+        insights-client --ansible-host={{ rhc_insights.ansible_host }} & wait
+      when:
+        - >-
+          "This host is registered" in __rhc_insights_status.stdout
+          or "Registered" in __rhc_insights_status.stdout
+      changed_when: true
+
+- name: Remove ansible host
+  when:
+    - rhc_insights.ansible_host is defined
+    - rhc_insights.ansible_host == __rhc_state_absent
+  block:
+    - name: Remove ansible host from insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^ansible_host="
+        state: absent
+    - name: Remove ansible host in inventory
+      shell: insights-client --ansible-host= & wait
+      register: __insights_ansible_host_removed
+      when:
+        - >-
+          "This host is registered" in __rhc_insights_status.stdout
+          or "Registered" in __rhc_insights_status.stdout
+      changed_when: true
+
 - name: Register insights-client
   shell: insights-client --register & wait
   when:
--- a/tests/tests_insights_ansible_host.yaml
+++ b/tests/tests_insights_ansible_host.yaml
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Insights ansible host test
+  hosts: all
+  gather_facts: false
+  become: true
+
+  tasks:
+    - name: Setup Insights
+      import_tasks: tasks/setup_insights.yml
+
+    - name: Test ansible_host
+      block:
+        - name: Add ansible_host and register insights
+          include_role:
+            name: linux-system-roles.rhc
+            public: true
+          vars:
+            rhc_auth:
+              login:
+                username: "{{ lsr_rhc_test_data.reg_username }}"
+                password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              ansible_host: host
+              remediation: absent
+              state: present
+            rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+            rhc_server:
+              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+              port: "{{ lsr_rhc_test_data.candlepin_port }}"
+              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+        - name: Check ansible_host is set to 'host' in config file
+          command:
+            grep -ixq "^ansible_host=host" {{ __rhc_insights_conf }}
+          changed_when: true
+
+        - name: Test ansible_host changed value after registration
+          block:
+            - name: Change ansible host to 'new-host'
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: new-host
+                  remediation: absent
+            - name: Check ansible_host is set to 'new-host' in config file
+              command:
+                grep -ixq "^ansible_host=new-host" {{ __rhc_insights_conf }}
+              changed_when: false
+            - name: Change ansible host to a null value (noop)
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: null
+                  remediation: absent
+            - name: Check ansible_host has not changed in config file
+              command:
+                grep -ixq "^ansible_host=new-host" {{ __rhc_insights_conf }}
+              changed_when: false
+
+        - name: Test ansible_host set to an absent value
+          block:
+            - name: Set ansible host to an absent value
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: {state: absent}
+                  remediation: absent
+            - name: Check ansible_host is not present in config file
+              lineinfile:
+                path: "{{ __rhc_insights_conf }}"
+                regexp: "^ansible_host="
+                state: absent
+              check_mode: true
+              register: __test_ansible_host_absent
+              failed_when: __test_ansible_host_absent.found
+
+        - name: Test ansible_host is removed during insights unregistration
+          block:
+            - name: Set ansible_host in config file
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: unreg-test
+                  remediation: absent
+            - name: Check ansible_host is set in config file
+              command:
+                grep -ixq "^ansible_host=unreg-test" {{ __rhc_insights_conf }}
+              changed_when: false
+            - name: Unregister insights to remove ansible_host in config file
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  state: absent
+            - name: Check ansible_host is not present in config file
+              lineinfile:
+                path: "{{ __rhc_insights_conf }}"
+                regexp: "^ansible_host="
+                state: absent
+              check_mode: true
+              register: __test_ansible_host_removed
+              failed_when: __test_ansible_host_removed.found
+
+      always:
+        - name: Unregister
+          include_role:
+            name: linux-system-roles.rhc
+          vars:
+            rhc_state: absent


tasks/insights-client.yml
```diff
@@ -36,6 +36,32 @@
     insertafter: "#auto_update"
     line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
 
+- name: Configure ansible host
```
richm, line 39: Not sure why you are using a `block` here - did you mean to use the `when` condition for the tasks in the block?

```yaml
  when: rhc_insights.ansible_host | d() != None
  block:
....
```
then you don't need the `when` for `ansible_host` below

tasks/insights-client.yml
```diff
@@ -36,6 +36,32 @@
...
+      shell: insights-client --ansible-host {{ host }} & wait
+      when:
+        - >-
+          "This host is registered" in __rhc_insights_status.stdout
+          or "Registered" in __rhc_insights_status.stdout
+        - >-
+          rhc_insights.ansible_host is defined
+          and rhc_insights.ansible_host != None
+      register: __rhc_insights_ansible_host_change_result
+      changed_when: true
```
richm, line 63: Can this be made idempotent?  That is, if `rhc_insights.ansible_host` is already the value of `insights-client --ansible-host`, it would be better to return `changed_when: false` in that case

tasks/insights-client.yml
DuckBoss: I made the block and forgot to pull out the `when` condition.
Thanks for the suggestion, it simplifies the tasks in the block a bit.

tasks/insights-client.yml
DuckBoss: I'm actually not sure about this because I don't think there's a way to check what the current `ansible_host` is from inventory. The `insights-client --ansible-host` command expects an input and updates inventory even if the same host name is provided as the one in inventory. Leaving the input param blank does not return the current host name in inventory afaik.

richm: ok - @ptoscano ^^^ ?

ptoscano: > I'm actually not sure about this because I don't think there's a way to check what the current `ansible_host` is from inventory.

Theoretically it could be done by querying the API manually, however that would add a gigantic amount of complications -> better not.

So I think the better way is to always set the hostname...

ptoscano (changes requested):
In addition to setting the hostname, we need to also be able to reset the hostname; since `rhc_insights.ansible_host` defaults to `null`, I think we should use the "standard" `{'state': absent'}` value for this (see the internal `__rhc_state_absent`).

tasks/insights-client.yml
ahitacat: I think this could be simplify. It is no need to use the `--ansible-host` parameter in the second task. Once you set the variable it will update the ansible-host name in inventory in the next collection.

Other way to achieve this is to use the parameter, and it will update in inventory. But it not keep tracks in the config file. That means the next collection will not use the new ansible-host name and will update to the one that is in configuration.

So I will update the config file, and make it idempotent checking with the ansible-host defined there.

tasks/insights-client.yml
```diff
@@ -36,6 +36,27 @@
     insertafter: "#auto_update"
     line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
 
+- name: Configure ansible host
+  when: rhc_insights.ansible_host | d() != None
```
ptoscano, line 40: ah, this check needs to be tweaked:
```yaml
  when:
    - not rhc_insights.ansible_host is none
    - rhc_insights.ansible_host != omit
```
Since `rhc_insights.ansible_host` is defined in `defaults/main.yml`, then it will be always defined.

The explicit check for `omit` is needed to avoid using the special value `omit` as real value; `omit` would be passed as role parameter in case there is no actual value for it -- e.g.:
```yaml
  rhc_insights:
    ansible_host: "{{ some_string if enable_host and some_string is defined else omit }}"
```

tasks/insights-client.yml
DuckBoss: How important is it for inventory to be immediately updated? If the ansible host name in inventory is not required to be immediately updated when using the role, then Alba's idea of updating only the config and waiting for the next collection would be a much simpler change.

Any further thoughts on this?

ptoscano: I think the general idea is that, in case you change something, that something is immediately reflected. I think it would be awkward otherwise that you change `ansible_host`, run the playbook/role, and there is no change for the system until the next check-in. Also, applying the change immediately is what is done already for tags -- see the `Do the collection if insights-client was registered` task.

So this would be yet another condition to trigger the manual collection.

tasks/insights-client.yml
```diff
@@ -36,6 +36,29 @@
...
 
+- name: Configure ansible host
+  when:
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != omit
+  block:
+    - name: Configure ansible host in config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^ansible_host"
```
ptoscano, line 47: put `=` in the regexp right after the key name, e.g.:
```        regexp: "^ansible_host="
```
this way, if there would ever be yet another option named `ansible_host<something>`, it would not be affected by this block.

tasks/insights-client.yml
DuckBoss: Agreed ^^^

I've pushed an update to simplify the yaml and added a condition to check for changes in the ansible host to trigger a manual  collection [[79e1bb8](https://github.com/linux-system-roles/rhc/pull/155/commits/79e1bb818dedfb1a0d91c9c26c76c9f94b39fa99)]. 
I hope that addresses everything in this thread.

DuckBoss:
I'll squash the commits after all the conversations are resolved -- thanks for the feedback so far everyone! :+1:

tasks/insights-client.yml
```diff
@@ -36,6 +36,17 @@
...
     line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
 
+- name: Configure ansible host
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^ansible_host"
+    insertafter: "#ansible_host"
+    line: ansible_host={{ rhc_insights.ansible_host }}
+  register: __insights_ansible_host_configured
+  when:
```
richm, line 46:
```suggestion
  when:
    - rhc_insights.ansible_host is defined
```
afaict it is possible for `rhc_insights.ansible_host` to be undefined, and that would cause an error

ptoscano: hm how so? since it's in the defaults, it should be always defined, shouldn't it? to me this looks like a mistake I did in the past, see #127

ptoscano (changes requested):
See also my previous comment on supporting unsetting the ansible host (`__rhc_state_absent`).

richm:
> hm how so? since it's in the defaults, it should be always defined, shouldn't it? to me this looks like a mistake I did in the past, see https://github.com/linux-system-roles/rhc/pull/127

I can call the role like this:
```yaml
- name: Run
  include_role:
    name: linux-system-roles.rhc
  vars:
    rhc_insights:
      autoupdate: true
```
in this case, `rhc_insights.ansible_host` will be undefined

ptoscano:
> > hm how so? since it's in the defaults, it should be always defined, shouldn't it? to me this looks like a mistake I did in the past, see #127
> 
> I can call the role like this:
> 
> ```yaml
> - name: Run
>   include_role:
>     name: linux-system-roles.rhc
>   vars:
>     rhc_insights:
>       autoupdate: true
> ```
> 
> in this case, `rhc_insights.ansible_host` will be undefined

Interesting, i would not have thought about it :thinking: TIL, thanks!

DuckBoss:
> > hm how so? since it's in the defaults, it should be always defined, shouldn't it? to me this looks like a mistake I did in the past, see #127
> 
> I can call the role like this:
> 
> ```yaml
> - name: Run
>   include_role:
>     name: linux-system-roles.rhc
>   vars:
>     rhc_insights:
>       autoupdate: true
> ```
> 
> in this case, `rhc_insights.ansible_host` will be undefined

Good catch! Thanks, I've pushed a commit to address this.

ahitacat:
@DuckBoss update also the README with this new parameter, so we can keep a kind of documentation for it :smile:

DuckBoss:
One issue I've noticed is that when the ansible host name is unset, it does not reset the ansible host in HBI to the system host name during the collection.
**Example:**
``` yaml
// Run playbook using role and set ansible_host to a custom value.
// This sets the new host in the cfg file and triggers a collection  to update HBI.
rhc_insights:
  ansible_host: test

// Run the playbook again, but unset the ansible_host value.
// ansible host name is removed from the cfg, but HBI remains as 'test'.
rhc_insights:
  ansible_host: {state: absent}
```
Are there any ways to resolve this without an update to insights-client?

DuckBoss:
I've made some larger changes in an effort to address the issues with updating HBI only when changes are detected and resetting HBI when the ansible host is absent in the config. 

One of the issues with resetting the ansible host in HBI is that a collection did not reset the HBI ansible host if the config had no value, and the only way to reset HBI is by passing an empty value to `--ansible-host` command. 
With the following changes, it's also not required to do a collection to update HBI.

I've listed some scenarios below:
#### Expected output when the ansible_host has not changed in the cfg:
In this situation, if the `ansible_host` param in the cfg is absent or already set to the `rhc_insights.ansible_host` value, then skip HBI updates.
 ```
 TASK [linux-system-roles.rhc : Update ansible host in insights-client config] ******************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Update ansible host in inventory] *******************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Remove ansible host from insights-client config] ****************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Remove ansible host in inventory] *******************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Register insights-client] ***************************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Do the collection if insights-client was registered] ************************************
skipping: [192.168.122.195]
 ```
#### Expected output when the ansible_host has changed in the cfg to a different value not already in the cfg:
```
TASK [linux-system-roles.rhc : Check ansible host in insights-client config] *******************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Update ansible host in insights-client config] ******************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Update ansible host in inventory] *******************************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Remove ansible host from insights-client config] ****************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Remove ansible host in inventory] *******************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Register insights-client] ***************************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Do the collection if insights-client was registered] ************************************
skipping: [192.168.122.195]
```
#### Expected output when the ansible_host is absent ( {state: absent} ):
```
TASK [linux-system-roles.rhc : Check ansible host in insights-client config] *******************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Update ansible host in insights-client config] ******************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Update ansible host in inventory] *******************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Remove ansible host from insights-client config] ****************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Remove ansible host in inventory] *******************************************************
changed: [192.168.122.195]

TASK [linux-system-roles.rhc : Register insights-client] ***************************************************************
skipping: [192.168.122.195]

TASK [linux-system-roles.rhc : Do the collection if insights-client was registered] ************************************
skipping: [192.168.122.195]
```

Looking for advice or suggestions about this approach and if there's a cleaner way to write the logic for checking the ansible host in the insights-client config.

ahitacat:
@DuckBoss I think you described all the possible used scenarios. However I have a concern where the change in HBI is made. All your tests are in a instance that is already registered, but what will happens if you try to set a display-name and update in inventory when that host doesn't exists there?

DuckBoss:
@ahitacat You're right. I've added back the conditions I had previously on `--ansible-host` commands to ensure the system is registered. (3395fd0)

ahitacat (changes requested):
I only have one last suggestion:

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,71 @@
...
+          vars:
+            rhc_insights:
+              ansible_host: null
+              remediation: absent
+
+        - name: Check ansible_host has not changed in config file
+          command:
+            grep -ixq "^ansible_host=new-test-host" {{ __rhc_insights_conf }}
+          changed_when: true
+
```
ahitacat, line 65: I think it could be a good idea to add a test for removing the display-host name, and another to check if after unregistration the display-name also change.

ptoscano: yes, see the unregistration task later on in the `always:` block

tasks/insights-client.yml
```diff
@@ -36,6 +36,59 @@
...
+- name: Check ansible host in insights-client config
+  when:
+    - rhc_insights.ansible_host is defined
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != omit
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^ansible_host="
+    state: "{{ 'present' if rhc_insights.ansible_host != __rhc_state_absent else 'absent' }}"
+    line: "{{ ('ansible_host='+rhc_insights.ansible_host) if rhc_insights.ansible_host != __rhc_state_absent }}"
```
ptoscano, line 48: sounds like you could rather add
```yaml
  - rhc_insights.ansible_host != __rhc_state_absent
```
as 3rd condition in the `when:` of this task: you don't need to check what exists if you need to drop it altogether

tasks/insights-client.yml
```diff
@@ -36,6 +36,59 @@
...
+  check_mode: true
+  register: __insights_ansible_host_exists
+
+- name: Configure ansible host
+  when:
+    - rhc_insights.ansible_host is defined
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != omit
+    - __insights_ansible_host_exists.changed
```
ptoscano, line 57: add
```yaml
  - rhc_insights.ansible_host != __rhc_state_absent
```
as 4th condition here (before checking/using `__insights_ansible_host_exists.changed`)

tasks/insights-client.yml
```diff
@@ -36,6 +36,59 @@
...
+        - rhc_insights.ansible_host != __rhc_state_absent
+        - >-
+          "This host is registered" in __rhc_insights_status.stdout
+          or "Registered" in __rhc_insights_status.stdout
+
+- name: Remove ansible host
+  when:
+    - rhc_insights.ansible_host is defined
+    - rhc_insights.ansible_host == __rhc_state_absent
+    - __insights_ansible_host_exists.changed
```
ptoscano, line 77: this condition can be dropped altogether: if we need to remove the `ansible_host`, then there is no need to check that it exists beforehand, `lineinfile` will do that

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: new-host
+                  remediation: absent
+            - name: Check ansible_host is set to 'new-host' in config file
+              command:
+                grep -ixq "^ansible_host=new-host" {{ __rhc_insights_conf }}
+              changed_when: true
```
ptoscano, line 52: the command here does not change things, so here it ought to be `changed_when: false`

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: null
+                  remediation: absent
+            - name: Check ansible_host has not changed in config file
+              command:
+                grep -ixq "^ansible_host=new-host" {{ __rhc_insights_conf }}
+              changed_when: true
```
ptoscano, line 63: ditto

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: unreg-test
+                  remediation: absent
+            - name: Check ansible_host is set in config file
+              command:
+                grep -ixq "^ansible_host=unreg-test" {{ __rhc_insights_conf }}
+              changed_when: true
```
ptoscano, line 95: ditto

defaults/main.yml
```diff
@@ -5,6 +5,7 @@ rhc_baseurl: null
 rhc_environments: []
 rhc_insights:
   autoupdate: true
+  ansible_host: null
```
ptoscano, line 8: since there are still changes to do, would you please move this `ansible_host` line one line up, so the list of items in `rhc_insights` is still alphasorted? :) (yes, nitpicks)

README.md
```diff
@@ -162,6 +162,21 @@ Whether the system is connected to Insights; valid values are `present`
 Whether the system automatically updates the dynamic configuration. It is
 enabled by default.
 
+```yaml
+    rhc_insights:
+      ansible_host: null
```
ptoscano, line 167: like the other examples that set string values, I'd specify an example of text/string

tasks/insights-client.yml
DuckBoss: I ran into some unexpected behavior with this suggestion.  Removing  the entire condition altogether  breaks the setting/updating of the `ansible_host` value in HBI since it results in the 2nd task in the block to always execute.

Did you mean to ask me to only remove this?
```
    - __insights_ansible_host_exists.changed
```
Removing only this condition works as expected with the other suggestions you've made.

README.md
```diff
@@ -162,6 +162,21 @@ Whether the system is connected to Insights; valid values are `present`
...
+      ansible_host: "example-host"
+```
+
+Configures the ansible host name with a custom value for the system record 
+in Host Based Inventory (HBI). This host name is used in playbooks by remediations.
+It is `null` by default and will use the system host name if not specified.
+
+Possible values of this variable:
+
+* `null` or an empty value: the ansible host name is not changed from the system host name.
```
ahitacat, line 172: I think this could be a bit confusing.
If you have previously set a ansible host name, and run a new playbook with this `null` parameter. I will expect the name to change to the system host name.

```suggestion
* `null` or an empty value: the ansible host name is not changed.
```

DuckBoss: Yeah this makes sense to me, thanks!

ahitacat (approved):
I have test it manually and this is working. So for me is OK to merge it.
Waiting for @ptoscano

ptoscano:
Thanks for the work so far, and the patience.

I gave the PR a number of tests, and it seems that it works well in all the scenarios I tried. I left a couple of minor nits, but other than that this PR seems ready to me.

Hence: please integrate the two small changes, rebase it, squash it, and ensure that the commit message and the PR title are nice and explain things well.

Thanks!

richm (approved):

ptoscano:
[citest]

ptoscano:
[citest bad]

ptoscano (approved):
The only test failure is in `tests_environments`, and it due to a regression in upstream Candlepin that broke the subscription-manager detection of environments. This appeared in the latest version, 4.4.3. I will create a PR to pin the candlepin container to the previous version for now.

Since that issue is unrelated to the changes of this PR, then I'm approving it & merging it.

Thanks for the work, @DuckBoss, good job!

defaults/main.yml
Glutexo: More nitpicks, please.

tasks/insights-client-unregistration.yml
```diff
@@ -11,3 +11,9 @@
       or "Registered" in __rhc_insights_status.stdout
   register: __rhc_insights_unregister_result
   changed_when: true
+
+- name: Reset ansible host in config
+  lineinfile:
```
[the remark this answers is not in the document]
richm: > Does Ansible have a plugin to handle INI files, so it’s not necessary to match the lines manually with a regexp? 

Yes and no - there is the community.general.ini_file module - but we really don't want the role to rely on 3rd party code unless absolutely necessary.

> Would this be sufficent with control characters that need to be escaped?

What is the issue?

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
     insertafter: "#auto_update"
     line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
 
+- name: Check ansible host in insights-client config
+  when:
+    - rhc_insights.ansible_host is defined
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != __rhc_state_absent
+    - rhc_insights.ansible_host != omit
```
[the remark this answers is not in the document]
richm: There is a difference between
```
    - rhc_insights.ansible_host != omit
```
and
```
    - rhc_insights.ansible_host != "omit"
```
the "magic" Ansible variable `omit` has a random, unique value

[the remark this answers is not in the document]
Glutexo: Oh, I see! I didn’t know that. Thank you! 🙇🏻‍♂️

I wondered how can I set _rhc_insights.ansible_host_ to this value. Found out I can do `{{ omit }}`. 😮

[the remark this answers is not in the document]
Glutexo: I am not exactly sure how is this different from _null_, but I guess that it’s an Ansible way of doing things and then we should support _omit_ too.

tasks/insights-client.yml
Glutexo: ++ for Pino’s „how so?“

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
...
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^ansible_host="
+    state: present
+    line: ansible_host={{ rhc_insights.ansible_host }}
+  check_mode: true
+  register: __insights_ansible_host_exists
```
Glutexo, line 53: The name _\_\_insights\_ansible\_host\_exists_ sounds misleading to me. It is more “matches the _rhc_insights.ansible_host_ value” rather than “exists”. But it may also mean “this exact ansible_host value exists”, in wich case it is ok.

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
...
+    - rhc_insights.ansible_host is defined
+    - rhc_insights.ansible_host == __rhc_state_absent
+  block:
+    - name: Remove ansible host from insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^ansible_host="
+        state: absent
+    - name: Remove ansible host in inventory
+      shell: insights-client --ansible-host= & wait
```
Glutexo, line 89: Running `insights-client --ansible-host=` really resets the Ansible Host value in the Inventory. Removing it from the config file, or setting it there to `ansible_host=`, however, doesn’t do so: after an upload, the value remains unchanged.

That means, the two changes do different things:

* The change to the configuration file causes Ansible Host to remain unchanged in the Inventory, regardless of what the stored value is.
* The CLI call causes the Ansible Host value to reset in the inventory.

README.md
```diff
@@ -158,6 +158,21 @@ Whether the system is connected to Insights; valid values are `present`
 Whether the system automatically updates the dynamic configuration. It is
 enabled by default.
 
+```yaml
+    rhc_insights:
+      ansible_host: "example-host"
+```
+
+Configures the ansible host name with a custom value for the system record
```
Glutexo, line 166:
```suggestion
Configures the Ansible host name with a custom value for the system record
```

README.md
```diff
@@ -158,6 +158,21 @@ Whether the system is connected to Insights; valid values are `present`
...
+in Host Based Inventory (HBI). This host name is used in playbooks by remediations.
```
Glutexo, line 167:
```suggestion
in Host Based Inventory (HBI). This host name is used in playbooks by Remediations.
```

README.md
```diff
@@ -158,6 +158,21 @@ Whether the system is connected to Insights; valid values are `present`
...
+It is `null` by default and will use the system host name if not specified.
+
+Possible values of this variable:
+
+* `null` or an empty value: the ansible host name is not changed.
```
Glutexo, line 172: What is an empty value? An empty string? The `omit` 
magic value?

```suggestion
* `null` or an empty value: the Ansible host name is not changed.
```

README.md
```diff
@@ -158,6 +158,21 @@ Whether the system is connected to Insights; valid values are `present`
...
+* `{state: absent}`: the ansible host name is unset in the insights-client config file and Host Based Inventory (HBI) is updated to use the system host name.
```
Glutexo, line 173:
```suggestion
* `{state: absent}`: the ansible host name is unset in the Insights Client config file and Host Based Inventory (HBI) is updated to use the system host name.
```

or

```suggestion
* `{state: absent}`: the ansible host name is unset in the insights-client.conf file and Host Based Inventory (HBI) is updated to use the system host name.
```

README.md
```diff
@@ -158,6 +158,21 @@ Whether the system is connected to Insights; valid values are `present`
...
+* any other string value: the ansible host name is changed in Host Based Inventory (HBI).
```
Glutexo, line 174:
```suggestion
* any other string value: the Ansible host name is changed in Host Based Inventory (HBI).
```

Glutexo (commented):
Some comments on the README.

README.md
richm: like this:
```yaml
    rhc_insights:
      ansible_host:
```
This is an "empty" value in YAML which is translated to a `null` - see https://yaml.org/spec/1.2.2/#24-tags "Example 2.21 Miscellaneous"

I say "empty" because, although it looks empty, that there is nothing there, in YAML this means to assign the value `null`

tasks/insights-client.yml
richm: > Is it ok that it is not possible to reset the Ansible Host value with the config file?

I'm not sure what you mean - do you mean "without using the `rhc` role"?

richm:
@Glutexo thank you for your interest in the `rhc` role.  Perhaps you could open a new PR for your suggested changes, rather than commenting on a closed PR?

tasks/insights-client-unregistration.yml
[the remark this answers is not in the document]
Glutexo: Oh. I didn’t kno that it’s 3rd party and we don’t want to rely on that.

For the escaping issue. I am not sure if there is any: I don’t know  by heard the details of how Python parses INI files. I was just afraid whether there aren’t some control characters that would cause that Python would read the value differently than hot it’s written without escaping.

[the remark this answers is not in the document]
richm: I suppose it is possible that someone could provide an `ansible_host` value with control characters, but that seems highly unlikely to me.

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
     insertafter: "#auto_update"
     line: auto_update={{ rhc_insights.autoupdate | d(true) | bool }}
 
+- name: Check ansible host in insights-client config
```
Glutexo, line 41:
```suggestion
- name: Check ansible host in Insights Client config
```
or
```suggestion
- name: Check ansible host in insights-client.conf
```

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
...
+
+- name: Configure ansible host
+  when:
+    - rhc_insights.ansible_host is defined
+    - not rhc_insights.ansible_host is none
+    - rhc_insights.ansible_host != omit
+    - rhc_insights.ansible_host != __rhc_state_absent
+    - __insights_ansible_host_exists.changed
+  block:
+    - name: Update ansible host in insights-client config
```
Glutexo, line 63:
```suggestion
    - name: Update ansible host in Insights Client config
```
or
```suggestion
    - name: Update ansible host in insights-client.conf
```

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
...
+    - rhc_insights.ansible_host == __rhc_state_absent
+  block:
+    - name: Remove ansible host from insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^ansible_host="
+        state: absent
+    - name: Remove ansible host in inventory
+      shell: insights-client --ansible-host= & wait
+      register: __insights_ansible_host_removed
```
Glutexo, line 90: This variable is not used anywhere. Is it necessary to register it? The previous task, too, doesn’t register a return variable.

```suggestion
```

richm:
@Glutexo thank you for your interest in the `rhc` role.  Please open a PR with your suggestions rather than make comments in a closed, merged PR.

tasks/insights-client.yml
Glutexo: Yes yes, using plain Insights Client CLI. I am aware that this is not the place to ask, it just came up on my mind.

tasks/insights-client.yml
Glutexo: I see! The “Update ansible host in inventory” task doesn’t need a return value?

Glutexo (commented):
Some comments and questions. In general I think the test playbook would deserve being split into multiple ones, each running on a pristine system.

README.md
Glutexo: I see. I knew about this feature of YAML, it just didn’t come to my mind.

Since the so-called empty value translates to _null_ on the YAML level, it may not be worth mentioning here: the value is still _null_ and the variations of the syntax are property of the YAML format, not Ansible or the role.

tasks/insights-client-unregistration.yml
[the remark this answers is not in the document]
Glutexo: I agree it’s unlikely. Maybe it’s not worth the effort. I would however expect from an official RHC role to be safe.

richm:
> Some comments and questions. In general I think the test playbook would deserve being split into multiple ones,

Not sure what you mean here.  Which test playbook should be split into multiple ones, and how would you do the split?

> each running on a pristine system.

We have that option in the test frameworks for tests that do not, or are unable to, clean up properly.

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+            rhc_server:
+              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+              port: "{{ lsr_rhc_test_data.candlepin_port }}"
+              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+        - name: Check ansible_host is set to 'host' in config file
+          command:
+            grep -ixq "^ansible_host=host" {{ __rhc_insights_conf }}
```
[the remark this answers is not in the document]
richm: > Can’t lineinfile be used here, rather than a grep command?

Yes, if there is a substantial benefit in terms of code readability and long term maintainability.  Since this is test code (as opposed to code in the `tasks/` runtime directory), I'm willing to allow a bit more leniency towards the preference of the original developer.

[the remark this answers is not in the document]
richm: > I’d rather consider single quoted string. Although there is no interpolation or control characters, it would convey a message of an exact string.

The general Ansible style is to only quote things that require quoting, so I would prefer no quotes at all.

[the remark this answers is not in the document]
richm: > Although the Insights Client config path is constant and doesn’t contain any special characters or spaces, paths can contain those. Paths may contain single quotes too, which can’t be properly escaped at all. It may make sense to rather store it in an enviornment variable.

Not sure what you mean by environment variable here - where would the value come from?  How would it be set for the task?  That seems unnecessarily complicated.  Note that Ansible has a built-in filter for quoting variables expanded in a shell command - `quote` - so this should correctly be `{{ __rhc_insights_conf | quote }}` - maybe with the environment variable comment you were getting at the fact that using any sort of single or double quoting is not guaranteed to work e.g. `'{{ __rhc_insights_conf }}'` will fail if `__rhc_insights_conf` (perversely) contains single quote characters.  See https://docs.ansible.com/ansible/2.9/user_guide/playbooks_filters.html#id8

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+              state: present
+            rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+            rhc_server:
+              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+              port: "{{ lsr_rhc_test_data.candlepin_port }}"
+              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+        - name: Check ansible_host is set to 'host' in config file
```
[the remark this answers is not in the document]
richm: You don't need to put values in task names.  In fact Ansible prefers not to have variables in task names because it makes it harder to find a task in the Ansible log output, and `grep` the source to figure out which task in the source the log output is from.  Ansible makes an allowance for having the variable at the end of the task name string for the rare occasions where it might be useful.

Since this is test code, I'm inclined to be more lenient.

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+          command:
+            grep -ixq "^ansible_host=host" {{ __rhc_insights_conf }}
+          changed_when: true
+
+        - name: Test ansible_host changed value after registration
+          block:
+            - name: Change ansible host to 'new-host'
+              include_role:
```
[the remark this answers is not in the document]
richm: > I wonder. Since the master task “Test ansible host” includes the role and sets _rhc_insights.state_ to _present_, it registers the system at its very start. This block, including the role again, triggers the registration on an already registered host?

No, because the role is smart enough to see that the system is already registered.  One important concept in Ansible is *idempotency* - running the role again with the same arguments should not change the state of the system, and Ansible should report no `"changed": true`.

> 
> If so, I’m afraid that the test scenario can be possibly affected by the tainted state.

I don't think so.

> In such case it would be necessary to have the test in its own playbook, which I think is generally a good idea when testing an Ansible role.

But not in this case.

> 
> In this case, the test should manually edit the config file by an _lineinfile_ task, then trigger the registration, and after that check that the value in the config file changed.

I don't think that's necessary in this case.

tests/tests_insights_ansible_host.yaml
```diff
@@ -0,0 +1,116 @@
...
+            - name: Check ansible_host is not present in config file
+              lineinfile:
+                path: "{{ __rhc_insights_conf }}"
+                regexp: "^ansible_host="
+                state: absent
+              check_mode: true
+              register: __test_ansible_host_removed
+              failed_when: __test_ansible_host_removed.found
+
+      always:
```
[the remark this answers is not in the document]
richm: > This is executed only once, not after each task in the block, right?

Right.

> What is the point of it being here, then? Just keeping the system clean for other playbooks?

That's one purpose.  Another purpose is to check the unregistration code (which is already checked elsewhere in this particular test).

>  I see there are no checks. What if the unregister role misbehaves?

In this particular test, the unregister code is already checked for misbehavior starting at line 102

Plus, we will get errors with subsequent tests if the `always:` block code did not clean up properly.

> Shouldn’t every playbooks run on a pristine system?

Yes and no.  One of the problems we have is performance.  If every test in every role must run in a new VM, then that's (number of tests) X (VM provisioning time) X (number of platforms) X (number of Ansible versions).  In addition there is the expense of VM provisioning.  So we make a compromise - roles that have tests that can properly clean up after the test run are run sequentially in the same VM (well, it is a bit more complicated than that) - and roles that cannot cleanup are able to use a single VM for each test.  This has worked very, very well for the past 5 years of system roles development.

tests/tests_insights_ansible_host.yaml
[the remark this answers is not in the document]
Glutexo: If there is `| quote`, which I didn’t know about, there is no need for something as complicated as an environment variable. 👍🏻  Cool!

tests/tests_insights_ansible_host.yaml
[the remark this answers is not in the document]
Glutexo: I see. The first task does the registration, and then, this one test that `ansible_host` is updated on an already registered host. Ok!

tests/tests_insights_ansible_host.yaml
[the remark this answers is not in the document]
Glutexo: I see. Thank you for such detailed explanation. It helped a lot! 🙇🏻‍♂️

Glutexo:
There are no tests that would verify that the Ansible Host has been updated in the Inventory: by querying the Inventory API with cURL and checking the values.

ptoscano:
> There are no tests that would verify that the Ansible Host has been updated in the Inventory: by querying the Inventory API with cURL and checking the values.

Since the tests here test the _role_, they limit themselves to ensure the right commands are run, that the config files are tweaked as needed, and so on. Testing that Inventory is updated as requested would effectively mean testing `insights-client` itself, which is out of the scope of the tests here.

Glutexo:
Good point, pino! 💪🏻 I don’t however see that the tests would verify that the commands are run. Only changes to the config file are tested.

tasks/insights-client.yml
```diff
@@ -38,6 +38,62 @@
...
+      when:
+        - >-
+          "This host is registered" in __rhc_insights_status.stdout
+          or "Registered" in __rhc_insights_status.stdout
+      changed_when: true
+
+- name: Remove ansible host
+  when:
+    - rhc_insights.ansible_host is defined
+    - rhc_insights.ansible_host == __rhc_state_absent
```
Glutexo, line 81:
```suggestion
    - rhc_insights.ansible_host == "" or rhc_insights.ansible_host == __rhc_state_absent
```

in case we wanted to use an empty string to remove the value. I don’t prefer this solution however.

ptoscano:
@Glutexo if you have concerns, please open a new PR or contact us directly, as @richm and me asked you already. A closed PR is not a personal task list. Thank you.

README.md
Glutexo, line 174: This is not true: an empty string value `""` removes the Ansible hostname value from the Inventory. The word “other” suggests that the “empty value” above may refer to an ampty string too, but that’s not true either and yields inconsistent results.

I‘d vote for removing the ambiguous wording “empty value” as `value:` is equivalent to `value: null` in YAML. See the conversation above.

```suggestion
* any non-empty string value: the ansible host name is changed in Host Based Inventory (HBI).
```

README.md
Glutexo: I found out that an empty string value yields inconsistent results:

- `ansible_host=` is put into the config file, causing the value not get updated on upload.
- The Client is called with `--ansible_host=`, removing the value from the Inventory.

Even though this “empty value” refers to the YAML syntax `value:`, it sounds ambiguous to me. I’d rather not describe features of YAML and be more explicit with empty strings.

```suggestion
* `null`: the ansible host name is not changed.
```

if we wanted to use an empty string to reset the value, which I don’t prefer, or

```suggestion
* `null` or an empty string: the ansible host name is not changed.
```

if we chose to consider an empty string an empty value, equivalent to `null` or `omit`.

It may be worth mentioning `omit` too.

```suggestion
* `null`, `omit`, or an empty string: the ansible host name is not changed.
```

README.md
Glutexo, line 173: If we decided that an empty string means the same as `{state: absent}`:

```suggestion
* `{state: absent}` or an empty string: the ansible host name is unset in the insights-client config file and Host Based Inventory (HBI) is updated to use the system host name.
```

I don’t prefer this solution though.

richm: For some history about why we use the seemingly unintuitive `{state: absent}` - see https://linux-system-roles.github.io/documentation/incremental_settings.html

Also, our usage of `omit` is problematic - `omit` should almost never be used except for as a direct argument to a module parameter - users should certainly not be setting role parameters to the value `omit` - I would prefer not to mention it.

PR #159, opened 2024-01-19
tests: temporarily pin Candlepin to 4.4.2

ptoscano:
The current latest release (4.4.3) introduces a regression that breaks the support for environments in subscription-manager. Since:
- we are not testing subscription-manager here, rather the rhc role
- the issue is already known by the Candlepin developers, and hopefully to be fixed in the next version

hence pin the Candlepin container to the latest version before the regression, i.e. 4.4.2. This way, we can keep testing changes in the rhc role.


ptoscano:
[citest]

ptoscano:
[citest bad]

richm (approved):

PR #161, opened 2024-02-01
Revert "tests: temporarily pin Candlepin to 4.4.2"

ptoscano:
Candlepin 4.4.4 fixes the regression that prevented environments to be used in subscription-manager. Hence, revert the container pinning, and keep using the latest version available.

This reverts commit 5e1bdc483baa335edf044e324fcd09d67b643359.


ptoscano:
[citest]

Issue #163, opened 2024-02-09
The test tests_insights_ansible_host.yaml is not executed


jharuda:
**Problem**
The Pull request [1] introduced the new test `tests_insights_ansible_host.yaml`.
The new test is not [2] executed by CI testing  because it has `yaml` suffix. CI expects the format `tests_*.yml` .

**Solution**
The fix is obvious. I suggest just to rename the test `tests_insights_ansible_host.yaml` to `tests_insights_ansible_host.yml`

[1] https://github.com/linux-system-roles/rhc/pull/155
[2] https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_rhc-161-57d4c57_CentOS-Stream-9_20240201-070654/artifacts/summary.html

PR #164, opened 2024-02-09
tests: rename tests_insights_ansible_host.yaml

jharuda:
It renames the test `tests_insights_ansible_host.yaml` to `tests_insights_ansible_host.yml`. It allows CI to execute this test.

See also: https://github.com/linux-system-roles/rhc/issues/163


richm (approved):
how did we miss this!?!?

PR #166, opened 2024-02-12
feat: Add a display name parameter

Glutexo:
Enhancement:
A new _rhc_insights.display_name_ parameter.

Reason:
The display name is a value managed by Insights Client. It is visible in the Host-Based Inventory and is used as the main visible name in the UI to identify a host and search for it.

Result:
The new _rhc_insights.display_name_ parameter sets the display name of the host. This value is written to the Insights Client config file and also set directly by a `--display-name` command. When marked as absent or on unregistration, the value is removed from the config file.

https://issues.redhat.com/browse/CCT-13

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,21 @@ Possible values of this variable:
 
 ```yaml
     rhc_insights:
+      display_name: "Example Host"
+```
+
+Configures the display name with a custom value for the system record in Host Based Inventory (HBI). This display name is only used to identify the host in the Inventory.
+It is `null` by default and will use the system host name if not specified.
+
+Possible values of this variable:
+
+* `null` or an empty string: the display name is not changed
+* any other string value: the display name is changed in Host Based Inventory (HBI).
+
+Note: If not set explicitly on registration, the display name is set to the hostname by default. It is not possible to automatically revert it to the hostname, but it can be set so manually.
+
+```yaml
+    rhc_insights:
       remediation: present
 ```
 
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -94,6 +94,43 @@
           or "Registered" in __rhc_insights_status.stdout
       changed_when: true
 
+- name: Check display name in insights-client config
+  when:
+    - rhc_insights.display_name is defined
+    - not rhc_insights.display_name is none
+    - rhc_insights.display_name != ""
+    - rhc_insights.display_name != omit
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^display_name="
+    state: present
+    line: display_name={{ rhc_insights.display_name }}
+  check_mode: true
+  register: __insights_display_name_exists
+
+- name: Configure display name
+  when:
+    - rhc_insights.display_name is defined
+    - not rhc_insights.display_name is none
+    - rhc_insights.display_name != ""
+    - rhc_insights.display_name != omit
+    - __insights_display_name_exists.changed
+  block:
+    - name: Update display name in insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^display_name="
+        insertafter: "#display_name="
+        line: display_name={{ rhc_insights.display_name }}
+    - name: Update display name in inventory
+      shell: >-
+        insights-client --display-name={{ rhc_insights.display_name }} & wait
+      when:
+        - >-
+          "This host is registered" in __rhc_insights_status.stdout
+          or "Registered" in __rhc_insights_status.stdout
+      changed_when: true
+
 - name: Register insights-client
   shell: insights-client --register & wait
   when:
--- a/tests/tests_insights_display_name.yml
+++ b/tests/tests_insights_display_name.yml
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Insights display name test
+  hosts: all
+  gather_facts: false
+  become: true
+
+  tasks:
+    - name: Setup Insights
+      import_tasks: tasks/setup_insights.yml
+
+    - name: Test display_name
+      block:
+        - name: Add display_name and register insights
+          include_role:
+            name: linux-system-roles.rhc
+            public: true
+          vars:
+            rhc_auth:
+              login:
+                username: "{{ lsr_rhc_test_data.reg_username }}"
+                password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              display_name: name
+              remediation: absent
+              state: present
+            rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+            rhc_server:
+              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+              port: "{{ lsr_rhc_test_data.candlepin_port }}"
+              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+        - name: Check display_name is set to 'name' in config file
+          command:
+            grep -ixq "^display_name=name" {{ __rhc_insights_conf }}
+          changed_when: true
+
+        - name: Test display_name changed value after registration
+          block:
+            - name: Change display name to 'new-name'
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  display_name: new-name
+                  remediation: absent
+            - name: Check display_name is set to 'new-name' in config file
+              command:
+                grep -ixq "^display_name=new-name" {{ __rhc_insights_conf }}
+              changed_when: false
+            - name: Change display name to a null value (noop)
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  display_name: null
+                  remediation: absent
+            - name: Check display_name has not changed in config file
+              command:
+                grep -ixq "^display_name=new-name" {{ __rhc_insights_conf }}
+              changed_when: false
+            - name: Change display name to an empty string (noop)
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  display_name: ""
+                  remediation: absent
+            - name: Check display_name has not changed in config file
+              command:
+                grep -ixq "^display_name=new-name" {{ __rhc_insights_conf }}
+              changed_when: false
+
+        - name: Test display_name is removed during insights unregistration
+          block:
+            - name: Set display_name in config file
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  display_name: unreg-test
+                  remediation: absent
+            - name: Check display_name is set in config file
+              command:
+                grep -ixq "^display_name=unreg-test" {{ __rhc_insights_conf }}
+              changed_when: false
+            - name: Unregister insights to remove display_name in config file
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  state: absent
+            - name: Check display_name is not present in config file
+              lineinfile:
+                path: "{{ __rhc_insights_conf }}"
+                regexp: "^display_name="
+                state: absent
+              check_mode: true
+              register: __test_display_name_removed
+              failed_when: __test_display_name_removed.found
+
+      always:
+        - name: Unregister
+          include_role:
+            name: linux-system-roles.rhc
+          vars:
+            rhc_state: absent
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,6 +6,7 @@ rhc_environments: []
 rhc_insights:
   ansible_host: null
   autoupdate: true
+  display_name: null
   remediation: present
   state: present
   tags: {}
--- a/tasks/insights-client-unregistration.yml
+++ b/tasks/insights-client-unregistration.yml
@@ -17,3 +17,9 @@
     path: "{{ __rhc_insights_conf }}"
     regexp: "^ansible_host="
     state: absent
+
+- name: Reset display name in config
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^display_name="
+    state: absent


tasks/insights-client.yml
```diff
@@ -94,6 +94,62 @@
...
+    - rhc_insights.display_name != __rhc_state_absent
+    - __insights_display_name_exists.changed
+  block:
+    - name: Update display name in insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^display_name="
+        insertafter: "#display_name="
+        line: display_name={{ rhc_insights.display_name }}
+    - name: Update display namein inventory
```
richm, line 125:
```suggestion
    - name: Update display name in inventory
```

README.md
```diff
@@ -173,6 +173,20 @@ Possible values of this variable:
 * `{state: absent}`: the ansible host name is unset in the insights-client config file and Host Based Inventory (HBI) is updated to use the system host name.
 * any other string value: the ansible host name is changed in Host Based Inventory (HBI).
 
+```yaml
+    rhc_insights:
+      display_name: "Example Host"
+```
+
+Configures the display name with a custom value for the system record in Host Based Inventory (HBI). This display name is only used to identify the host in the Invenotry.
```
ptoscano, line 181: typo "Invenotry"

ptoscano (commented):
seems a good start; please don't forget a test for it

README.md
Glutexo: If the only comment on the description is a typo, it’s good news! 🥳

tasks/insights-client.yml
Glutexo: Also probably Inventory should be written with a capital I. But keeping that lowercase for consistency with the Ansible Host tasks descriptions.

Glutexo:
Fixed the typos. Going to add the tests.

Glutexo:
I added the tests, so I am opening the pull request for a review.

DuckBoss:
I had a quick look over -- make sure the tests are referencing `display_name` instead of `ansible_host`. I know the tests are near-identical to `ansible_host` tests but I'm assuming leaving the fields as `ansible_host` was unintentional.

I can test these changes shortly after you push the changes :+1:

DuckBoss (changes requested):
I had a quick look over -- make sure the tests are referencing `display_name` instead of `ansible_host`. I know the tests are near-identical to `ansible_host` tests but I'm assuming leaving the fields as `ansible_host` was unintentional.

I can test these changes shortly after you push the changes :+1:

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+- name: Insights display name test
+  hosts: all
+  gather_facts: false
+  become: true
+
+  tasks:
+    - name: Setup Insights
+      import_tasks: tasks/setup_insights.yml
+
+    - name: Test ansible_host
```
richm, line 12:
```suggestion
    - name: Test display_name
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+      block:
+        - name: Add ansible_host and register insights
```
richm, line 14:
```suggestion
        - name: Add display_name and register insights
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+          include_role:
+            name: linux-system-roles.rhc
+            public: true
+          vars:
+            rhc_auth:
+              login:
+                username: "{{ lsr_rhc_test_data.reg_username }}"
+                password: "{{ lsr_rhc_test_data.reg_password }}"
+            rhc_insights:
+              ansible_host: host
```
richm, line 24:
```suggestion
              display_name: friendly_hostname
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              state: present
+            rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+            rhc_server:
+              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+              port: "{{ lsr_rhc_test_data.candlepin_port }}"
+              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+
+        - name: Check ansible_host is set to 'host' in config file
```
richm, line 35:
```suggestion
        - name: Check display_name is set to 'friendly_hostname' in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+          command:
+            grep -ixq "^ansible_host=host" {{ __rhc_insights_conf }}
```
richm, line 37:
```suggestion
            grep -ixq "^display_name=friendly_hostname" {{ __rhc_insights_conf }}
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+          changed_when: true
+
+        - name: Test ansible_host changed value after registration
```
richm, line 40:
```suggestion
        - name: Test display_name changed value after registration
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+          block:
+            - name: Change ansible host to 'new-host'
```
richm, line 42:
```suggestion
            - name: Change display_name to 'new-host'
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: new-host
```
richm, line 47:
```suggestion
                  display_name: new-host
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+                  remediation: absent
+            - name: Check ansible_host is set to 'new-host' in config file
```
richm, line 49:
```suggestion
            - name: Check display_name is set to 'new-host' in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              command:
+                grep -ixq "^ansible_host=new-host" {{ __rhc_insights_conf }}
```
richm, line 51:
```suggestion
                grep -ixq "^display_name=new-host" {{ __rhc_insights_conf }}
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              changed_when: false
+            - name: Change ansible host to a null value (noop)
```
richm, line 53:
```suggestion
            - name: Change display_name to a null value (noop)
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: null
```
richm, line 58:
```suggestion
                  display_name: null
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+                  remediation: absent
+            - name: Check ansible_host has not changed in config file
```
richm, line 60:
```suggestion
            - name: Check display_name has not changed in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              command:
+                grep -ixq "^ansible_host=new-host" {{ __rhc_insights_conf }}
```
richm, line 62:
```suggestion
                grep -ixq "^display_name=new-host" {{ __rhc_insights_conf }}
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              changed_when: false
+
+        - name: Test ansible_host set to an absent value
```
richm, line 65:
```suggestion
        - name: Test display_name set to an absent value
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+          block:
+            - name: Set ansible host to an absent value
```
richm, line 67:
```suggestion
            - name: Set display_name to an absent value
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: {state: absent}
```
richm, line 72:
```suggestion
                  display_name: {state: absent}
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+                  remediation: absent
+            - name: Check ansible_host is not present in config file
```
richm, line 74:
```suggestion
            - name: Check display_name is not present in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              lineinfile:
+                path: "{{ __rhc_insights_conf }}"
+                regexp: "^ansible_host="
```
richm, line 77:
```suggestion
                regexp: "^display_name="
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+                state: absent
+              check_mode: true
+              register: __test_ansible_host_absent
```
richm, line 80:
```suggestion
              register: __test_display_name_absent
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              failed_when: __test_ansible_host_absent.found
```
richm, line 81:
```suggestion
              failed_when: __test_display_name_absent.found
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+
+        - name: Test ansible_host is removed during insights unregistration
```
richm, line 83:
```suggestion
        - name: Test display_name is removed during insights unregistration
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+          block:
+            - name: Set ansible_host in config file
```
richm, line 85:
```suggestion
            - name: Set display_name in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  ansible_host: unreg-test
```
richm, line 90:
```suggestion
                  display_name: unreg-test
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+                  remediation: absent
+            - name: Check ansible_host is set in config file
```
richm, line 92:
```suggestion
            - name: Check display_name is set in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              command:
+                grep -ixq "^ansible_host=unreg-test" {{ __rhc_insights_conf }}
```
richm, line 94:
```suggestion
                grep -ixq "^display_name=unreg-test" {{ __rhc_insights_conf }}
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              changed_when: false
+            - name: Unregister insights to remove ansible_host in config file
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_insights:
+                  state: absent
+            - name: Check ansible_host is not present in config file
```
richm, line 102:
```suggestion
            - name: Check display_name is not present in config file
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              lineinfile:
+                path: "{{ __rhc_insights_conf }}"
+                regexp: "^ansible_host="
```
richm, line 105:
```suggestion
                regexp: "^display_name="
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+                state: absent
+              check_mode: true
+              register: __test_ansible_host_removed
```
richm, line 108:
```suggestion
              register: __test_display_name_removed
```

tests/tests_insights_display_name.yml
```diff
@@ -0,0 +1,116 @@
...
+              failed_when: __test_ansible_host_removed.found
```
richm, line 109:
```suggestion
              failed_when: __test_display_name_removed.found
```

Glutexo:
🤦🏻‍♂️ I did those changes already, but of couse I didn’t add them to Git before committing. Thank you for spotting that! Pushed the fixed test file.

tests/tests_insights_display_name.yml
Glutexo: friendly_hostname 😊

Glutexo:
Filled in the pull request description.

richm:
[citest]

tasks/insights-client.yml
```diff
@@ -94,6 +94,62 @@
...
+    - rhc_insights.display_name is defined
+    - rhc_insights.display_name == __rhc_state_absent
+  block:
+    - name: Remove display name from insights-client config
+      lineinfile:
+        path: "{{ __rhc_insights_conf }}"
+        regexp: "^display_name="
+        state: absent
+    - name: Remove display name in inventory
+      shell: insights-client --display-name= & wait
```
DuckBoss, line 145: Resetting the display name using this command doesn't seem to be working as expected in insights-client. It just starts another insights collection instead of reverting the display-name back to the system name which causes this to silently fail.

I'll have a look at the insights-client code that handles this, because it looks like it doesn't handle empty values being passed through like the `ansible-host` command.

DuckBoss (changes requested):

richm:
> The implementation seems mostly OK, and the test passes when running it against Insights.
> 
> The problem I see is when unsetting the display name (i.e. `{state: absent}`): whereas the configuration is properly updated, the display name in Inventory is not changed. In particular, it looks like `insights-client --display-name=` does not do much, as it is considered `None` by insights-core, and thus it is a no-op.
> 
> I see few possibilities for this:
> 
>     1. fix insights-core to allow an empty string for the display name as synonym for "use the system hostname"
> 
>     2. manually pass an hostname here in the role by getting it manually; the logic in insights-core is IMHO not easy to duplicate manually [1], so it would be either
>        
>        * unconditionally use something like `hostname -f`
>        * put the logic below in a small private module for this role
> 
>     3. "disallow" unsetting the display name: an empty string would behave as if `display_name:` was not passed (i.e. not changing it), and document this situation
> 
> 
> [1] https://github.com/RedHatInsights/insights-core/blob/396355192c229982f5cb55288b92c21a82151240/insights/util/hostname.py
> 
> Personally, I'd go for (2a) or (3).
> 
> @Glutexo @richm What do you think?

2. seems the best option to me

tasks/insights-client.yml
DuckBoss: Oops didn't notice Pino already posted about this issue!

ptoscano:
>     2. seems the best option to me

(2) as in "the simple `hostname` way", or the small helper private module? (I put each as possible option, not both together)

richm:
> > ```
> > 2. seems the best option to me
> > ```
> 
> (2) as in "the simple `hostname` way", or the small helper private module? (I put each as possible option, not both together)

As in the simple `hostname` way

richm:
> @ptoscano I prefer option (3) and updating insights-core afterwards to have similar functionality to `--ansible-host` when given empty values. Personally, id prefer not to add custom logic in the role to handle this with option (2) in this case.

3 is also fine with me

ptoscano:
> @ptoscano I prefer option (3) and updating insights-core afterwards to have similar functionality to `--ansible-host` when given empty values. Personally, id prefer not to add custom logic in the role to handle this with option (2) in this case.

Makes sense to me; let's go for this then.

Glutexo:
Oh, I didn’t know `insights-client --display-name=` does nothing. That’s exatly the think that I would have caught during proper manual testing. Nice catch! ⚾️

I completely agree with @DuckBoss here on going with way ⑶, treating an empty string as a _null_ value; and of course documenting it. I believe that the logic of determining a hostname (way ⑵) doesn’t belong to the role: it’s reponsibility of the Client, which doesn’t support this (yet).

It makes sense to me to add the feature to the Client, so the behavior is equivalent to that of `--ansible_host`, which would be reasonable even without the RHC role. Once it’s implemented, we’d revisit the role and make it use the new feature. So, way ⑴, but as a separate step.

We have a consensus of alll four of us on the way to go. Going to do that.

Glutexo:
I removed the subtask trying to unset the display name by running `insights-client display_name=` and updated the README. The changes are pushed to this PR.

richm:
[citest]

richm:
lgtm - @ptoscano @DuckBoss ?

DuckBoss (approved):
LGTM, let's revisit wrt `--display-name` after changes to insights-client.

ptoscano (changes requested):
> * `{state: absent}`: the display name is unset in the insights-client config file; please note that Host Based Inventory is not updated

I don't think this is a good approach: the user will see that the insights-client configuration was changed (good), but not what is in Inventory (bad), leading the users to think that their operation succeeded. If it is not possible to "unset" (i.e. reset to the hostname) the display name, then the role ought to not allow the users to do this.

Hence:
- drop the handling of `{state: absent}` altogether
- handle empty values as `null`, documenting this behaviour as limitation in the underlying tooling (`insights-client`)

ptoscano (changes requested):
@Glutexo not sure why I'm requesting to review again; see my previous comment on 'state: absent'

Glutexo:
I spotted an issue when `display_name` is set to an empty string:

- `display_name=` gets written into the config file, causing the display name not to be updated. It is equivalent to a missing value in the file.
- The Client is run with `--display_name=`, which is ignored and causes a regular connection to run. That is an unwanted and resource-heavy process.

It is the same problem as I found in https://github.com/linux-system-roles/rhc/pull/155#pullrequestreview-1882483685. Since this pull request already has an approval from Jason, it may make sense to merge it as it is and fix the bug for both parameters at the same time in a follow-up pull request. If you think it’s worth including in this one, I‘ll push the fix.

Glutexo:
@ptoscano

The `{state: absent}` is however the only way how to remove the value from the config file, causing it at least not to be overwritten on collection. It is not an ideal solution, but as long as it’s documented it at least provides the user with an additional option.

Moreover, the non-ideal state that may cause confusion to the uses (although the real behavior is documented) would only last until the Client is patched to support display name removal. That is true for the absence of `{state: absent}` too, though.

Still, I’d be more inclined to offer a way to clean the configuration, when there is one to add it.

What do you think, Pino?

ptoscano:
> The `{state: absent}` is however the only way how to remove the value from the config file, causing it at least not to be overwritten on collection.
> [...]
> Still, I’d be more inclined to offer a way to clean the configuration, when there is one to add it.

I'm not sure what would be the point of removing the value from the config file, if that will not reflect what is the state in Inventory.

Please see my previous comment which explains the situation as I see it: https://github.com/linux-system-roles/rhc/pull/166#pullrequestreview-1883619178

> Moreover, the non-ideal state that may cause confusion to the user (although the real behavior is documented) would only last until the Client is patched to support display name removal. That is true for the absence (😃) of `{state: absent}` too, though.

Yes, and I do not think the role ought to do operations that do not actually work. When `insights-client` (or actually `insights-core`) will be changed to allow resetting/unsetting the display name, then the role may be adapted to make use of that.

Glutexo:
Since Pino requested changes on the pull request, I believe I can fix the empty string issue in this pull request. Pushed a new commit that causes the empty string to be treated as an empty value: the display name is ignored in that case.

Glutexo:
Pushed a patch removing the `{state: absent}` value. If a user still tries to provide such value, it is converted to a string by Python (to `{'state': 'absent'}`, which causes error. I believe this is not something we need to cover, as this happens to other string values too.

README.md
```diff
@@ -173,6 +173,19 @@ Possible values of this variable:
...
+      display_name: "Example Host"
+```
+
+Configures the display name with a custom value for the system record in Host Based Inventory (HBI). This display name is only used to identify the host in the Inventory.
+It is `null` by default and will use the system host name if not specified.
+
+Possible values of this variable:
+
+* `null` or an empty string: the display name is not changed
+* any other string value: the display name is changed in Host Based Inventory (HBI).
```
richm, line 187: Do we need a `NOTE: Once set, the display name cannot be removed or reverted back to the original value.  You can only set a new value.`
?

ptoscano: Strictly speaking, there is always a display name set: in particular, the default one that `insights-client` sets (if not specified otherwise) is the hostname.

I would either write something slightly more verbose along the lines of

> Note: `insights-client` sets the default display name on registration as the hostname of the system, in case it is not explicitly configured. When changing the display name manually using this variable, it is not possible to "revert it" automatically to be the hostname, and thus it must be done manually if wanted.

... or nothing at all.

ptoscano:
> The role is not primarily intended to edit the Client config file. It uses the Client as a tool to perform changes in the Inventory and changes the config file so the configuration is consistent with the new state and doesn’t get reverted on the next collection.

Yes and no.

Take a step back: a role/playbook/module/etc is nothing more than a glorified/structured way to do operations and check their results. Whatever is done with Ansible, it can be run manually to get the same result.

In case of this role, its purpose is to create and maintain the connectivity of systems to Red Hat or Satellite, Insights included. If to do so some configuration files must be edited, then they are edited; if some command must be run, they are run; and so on.

In the specific case of this option, we need to ensure that `insights-client` will keep reporting the configured display name to the connected Inventory. Since querying manually Inventory would sort of "go around" the purpose of `insights-client`, then we tell `insights-client` which display name to use by writing it in the config file, and then running `insights-client` as needed to ensure Inventory is up-to-date with the wanted configuration.

ptoscano (changes requested):
The changes seem fine to me now, thanks!

The only thing to check would be what Rich mentioned (and I added some notes) on the documentation about the lack of "unset"; let's include what I wrote (or something along those lines), if that looks OF for you.

Since the time for merging is closer, please rebase and squash the commits into one with a proper commit message for it.

README.md
Glutexo: Is is true that it’s the `insights-client` that sets the display name to the hostname, or is it sent empty and set automatically in the cloud?

Glutexo: Answer: It’s not the Client that sets the default display name. If not set, it doesn’t send it at all:

- no _display_name_ field in the canonical facts JSON;
- no _display_name_ file in the archive.

That means that the system gets its display name when processed in the Platform, not on the machine.

Glutexo:
I added a note on the default display name value to the documentation. I found the suggested wording a bit too verbose, so I tried to shorten it a bit.

ptoscano:
> please rebase and squash the commits into one with a proper commit message for it.

this still needs to be done

Glutexo:
Squashed the branch into a single commit and force-pushed.

richm:
[citest]

Glutexo:
I realized that I added the empty string condition to fix a buggy behavior, but without a test.

I added a test verifying that setting _rhc_insights.display_name_ to an empty string doesn’t change the value in the config file, just as a _null_ doesn’t. Without the added condition, the test would fail, because the empty string would change the line in the config file to `display_name=`.

I didn’t test calling the `insights-client`, because that is not tested at all by any of the existing tests.

ptoscano:
[citest]

Glutexo:
From what I found that the Insights Client doesn’t send an empty display name, I am a bit afraid that adding support for the Client to reset the value in the Inventory would require a change on the Platform side. I will however verify, what is the actual behavior of sending various empty values of the display name.

ptoscano (approved):
LGTM now -- thanks for the work on this!

PR #177, opened 2024-05-03
feat: Add support to configure parameters form insights-client.conf

Tronde:
Enhancement: Make most of the parameters in `/etc/insights-client/insights-client.conf` configurable/manageable by this role.

Reason: This enables users to configure the `insights-client` parameters at scale.

Result: Currently only the parameters `autoupdate`, `ansible_host`, `display_name`, `remediation`, `tags`, and `rhc_proxy` are supported by this role. This PR adds support for the following parameters:

- `loglevel`
- `auto_config`
- `authmethod`
- `username`
- `password`
- `base_url`
- `obfuscate`
- `obfuscate_hostname`
- `redaction_file`
- `content_redaction_file`
- `tags_file`

[https://issues.redhat.com/browse/RHEL-35245](https://issues.redhat.com/browse/RHEL-35245)

Additional information:

- As I'm not a software developer and this is my first PR for this role, please bear with me if I have missed something or did something wrong
- I followed the [contribution guide](https://linux-system-roles.github.io/contribute.html) to my best judgement
- To add the new features I copied, pasted, and modified existing tasks so my contribution should align well with the existing code
- Advice on how to improve are appreciated

### DIFF
--- a/tasks/insights-client.yml
+++ b/tasks/insights-client.yml
@@ -89,7 +89,6 @@
   when:
     - rhc_insights.baseurl is defined
     - not rhc_insights.baseurl is none
-    - not rhc_insights.baseurl is null
     - rhc_insights.baseurl != ""
     - rhc_insights.baseurl != __rhc_state_absent
     - rhc_insights.baseurl != omit


README.md
```diff
@@ -173,6 +192,28 @@ Possible values of this variable:
...
+    rhc_inights:
+      baseurl: "cert-api.access.redhat.com:443/r/insights"
+```
+
+Configures the Base URL for the Insights API.
+Defaults to "cert-api.access.redhat.com:443/r/insights".
+
+```yaml
+    rhc_insights:
+      redaction_file: "/etc/insights-client/file-redaction.yaml"
```
richm, line 205: Are these file paths the defaults?  What are the reasons someone would want to change these from the default values?

tasks/insights-client.yml
```diff
@@ -38,6 +38,137 @@
...
+    - rhc_insights.password is defined
+    - not rhc_insights.password is none
+    - rhc_insights.password != ""
+    - rhc_insights.password != omit
+  lineinfile:
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^password="
+    state: present
+    line: password={{ rhc_insights.password }}
+  check_mode: true
```
richm, line 83: This will likely need `no_log: true` in order to prevent the cleartext password from being leaked to the ansible log.

README.md
richm: How will users provide the contents of file-redaction.yaml and file-content-redaction.yaml?

tasks/insights-client.yml
richm: In fact - every task that can display the contents of `rhc_insights` will now need `no_log: true` in order to prevent leaking the cleartext `password`

To verify - run the playbook with `-vv` - then `grep` the logs/output for the given password.

README.md
Tronde: Yes, these file paths are the defaults. To be honest I don't know why someone would like to change these. I thought when making parameters configurable with this role it might just include these as well.

On the question how to provide the content I can think of two options (for both the code is currently missing):

- Template these files using ˋansible.builtin.templateˋ
- Have them stored somewhere reachable for the controller to copy them to the managed node by using ˋansible.builtin.copyˋ

Using Jinja2 templates to create these files would be more flexible but also more complex.

tasks/insights-client.yml
Tronde: I see your point. I'm going to move authentication related variables to another dictionary. Then setting ˋno_log: trueˋ for all tasks accessing ˋrhc_insightsˋ won't be necessary anymore.

Thank you for your comments. :-)

README.md
richm: > Yes, these file paths are the defaults. To be honest I don't know why someone would like to change these. I thought when making parameters configurable with this role it might just include these as well.

Ok.  I think the more likely use case is someone wanting to customize the settings in file-redaction.yaml and file-content-redaction.yaml - not sure if @ptoscano had plans for that.

> 
> On the question how to provide the content I can think of two options (for both the code is currently missing):
> 
>     * Template these files using ˋansible.builtin.templateˋ
> 
>     * Have them stored somewhere reachable for the controller to copy them to the managed node by using ˋansible.builtin.copyˋ

* provide the contents inline.

A few of our roles do something similar - for example, the podman role allows the user to provide quadlet definitions either in a file (using `copy`), a template (using `template`), or inline - see https://github.com/linux-system-roles/podman/?tab=readme-ov-file#podman_quadlet_specs

> 
> 
> Using Jinja2 templates to create these files would be more flexible but also more complex.

Yes. and we can support one of the methods at first then support the others later.  This is why I think we need to identify the use cases (perhaps @ptoscano has already done this).

Tronde:
With commit [e2395bc](https://github.com/linux-system-roles/rhc/pull/177/commits/e2395bc95cb586f92a1aa49a7bccd408e2471f60) I implemented the configuration to create `file-redaction.yaml` and `file-content-redaction.yaml`.

I guess that's it. I wait for the review by @ptoscano. :-)

README.md
```diff
@@ -173,6 +192,14 @@ Possible values of this variable:
 * `{state: absent}`: the ansible host name is unset in the insights-client config file and Host Based Inventory (HBI) is updated to use the system host name.
 * any other string value: the ansible host name is changed in Host Based Inventory (HBI).
 
+```yaml
+    rhc_inights:
```
richm, line 196:
```suggestion
    rhc_insights:
```

README.md
```diff
@@ -188,6 +215,56 @@ Possible values of this variable:
...
 
+```yaml
+    rhc_insights:
+      file_redaction:
+        commands: []
+        files: []
+        components: []
+```
+
+Specify lists of commands, files, and components to obmit from output
```
richm, line 226:
```suggestion
Specify lists of commands, files, and components to omit from output
```

README.md
```diff
@@ -188,6 +215,56 @@ Possible values of this variable:
...
+        files: []
+        components: []
+      file_content_redaction:
+        keywords: []
+        patterns: []
+        regex_patterns: []
+```
+
+These are optional.
+Specify lists of commands, files, components, keywords and patterns to obmit from output.
```
richm, line 241:
```suggestion
Specify lists of commands, files, components, keywords and patterns to omit from output.
```

README.md
```diff
@@ -188,6 +215,56 @@ Possible values of this variable:
...
+NOTE: You cannot mix plain string matching and regular expression matching.
+For more informaition on this topic read: [YAML-style denylist configuration for Red Hat Insights Client](https://access.redhat.com/articles/4511681).
```
richm, line 243:
```suggestion
For more information on this topic read: [YAML-style denylist configuration for Red Hat Insights Client](https://access.redhat.com/articles/4511681).
```

README.md
```diff
@@ -188,6 +215,56 @@ Possible values of this variable:
...
+
+Configures the log level; valid options are DEBUG (the default), INFO, WARNING, ERROR, CRITICAL.
+
+```yaml
+    rhc_insights:
+      obfuscate: false
+```
+
+Configures IP address obfuscation; valid values are `false` (the default),
+and `false`.
```
richm, line 258:
```suggestion
and `true`.
```

README.md
```diff
@@ -188,6 +215,56 @@ Possible values of this variable:
...
+
+```yaml
+    rhc_insights:
+      obfuscate_hostname: false
+```
+
+Configures hostname obfuscation; valid values are `false` (the default),
+and `false`. Requires `obfuscate: true`.
```
richm, line 266:
```suggestion
and `true`. Requires `obfuscate: true`.
```

README.md
```diff
@@ -173,6 +192,14 @@ Possible values of this variable:
 * `{state: absent}`: the ansible host name is unset in the insights-client config file and Host Based Inventory (HBI) is updated to use the system host name.
 * any other string value: the ansible host name is changed in Host Based Inventory (HBI).
 
+```yaml
+    rhc_inights:
+      baseurl: "cert-api.access.redhat.com:443/r/insights"
+```
+
+Configures the Base URL for the Insights API.
+Defaults to "cert-api.access.redhat.com:443/r/insights".
```
richm, line 201: @ptoscano Is this value built-in to the insights client?  where does it come from?

defaults/main.yml
```diff
@@ -5,11 +5,28 @@ rhc_baseurl: null
 rhc_environments: []
 rhc_insights:
   ansible_host: null
+  autoconfig: true
   autoupdate: true
+  baseurl: "cert-api.access.redhat.com:443/r/insights"
```
richm, line 10: I would rather not have an explicit value like this, I would rather have a value which means "if `baseurl` is unset, use whatever is the default value built-in to insights client".

README.md
Tronde: It's the default value from `/etc/insights-client/insights-client.conf`:

```
# Base URL for the Insights API
#base_url=cert-api.access.redhat.com:443/r/insights
```

defaults/main.yml
Tronde: I put the value in `defaults/main.yml` as I thought it is a good idea to set the default value from the `/etc/insights-client/insights-client.conf` as a default value for the role as well.

I can see that this could cause some issues if there are different defaults for different releases or when the default in insights-client change a change to the role would be necessary, too. I can change that.

README.md
Tronde: Fixed in commit [ff0199b](https://github.com/linux-system-roles/rhc/pull/177/commits/ff0199b1f95e5490a32844b1530f5b49f3a0167f)

README.md
Tronde: Fixed in commit [ff0199b](https://github.com/linux-system-roles/rhc/pull/177/commits/ff0199b1f95e5490a32844b1530f5b49f3a0167f)

README.md
Tronde: Fixed in commit [ff0199b](https://github.com/linux-system-roles/rhc/pull/177/commits/ff0199b1f95e5490a32844b1530f5b49f3a0167f)

README.md
Tronde: Fixed in commit [ff0199b](https://github.com/linux-system-roles/rhc/pull/177/commits/ff0199b1f95e5490a32844b1530f5b49f3a0167f)

README.md
Tronde: Fixed in commit [ff0199b](https://github.com/linux-system-roles/rhc/pull/177/commits/ff0199b1f95e5490a32844b1530f5b49f3a0167f)

README.md
Tronde: Fixed in commit [ff0199b](https://github.com/linux-system-roles/rhc/pull/177/commits/ff0199b1f95e5490a32844b1530f5b49f3a0167f)

defaults/main.yml
Tronde: But what's the better approach in general? To put default values from client configs into `default/main.yml` or not? Is there some rule of thumb or does it depend on every single case?

When you can give me some advice I will consider it for future contributions.

richm: > But what's the better approach in general? To put default values from client configs into default/main.yml or not? Is there some rule of thumb or does it depend on every single case?

There is no hard and fast rule.  I haven't looked at https://github.com/redhat-cop/automation-good-practices to see if this is covered.  And maybe it is just my opinion, and others will probably disagree with me.

The problem with this particular variable is that now the default value is set it in two different places, and it puts a burden on the role maintainer to constantly check if the value has changed in `/etc/insights-client/insights-client.conf` in order to keep the role in sync with the insights default.  This is not good programming practice.  In addition, if the value in the role gets out of sync, then the failure case will be mysterious and hard to debug.

In the case of other values in defaults/main.yml, these reasons do not apply.

Tronde: I see your point and agree. I'm going to change the value in `defaults/main.yml` to `baseurl: null` and edit the corresponding task that it will only run when another value was provided.

Tronde: Fixed in commit [1600b96](https://github.com/linux-system-roles/rhc/pull/177/commits/1600b964e8a3ebc1772557532a5c80db72866d34).

tasks/insights-client.yml
```diff
@@ -38,6 +38,96 @@
...
+    path: "{{ __rhc_insights_conf }}"
+    regexp: "^authmethod="
+    state: present
+    line: authmethod={{ rhc_insights_auth.authmethod }}
+  check_mode: true
+  no_log: true
+
+- name: Configure username for authmethod BASIC
+  when:
+    - rhc_insights_auth.authmethod = "BASIC"
```
richm, line 58:
```suggestion
    - rhc_insights_auth.authmethod is defined
    - rhc_insights_auth.authmethod = "BASIC"
```

tasks/insights-client.yml
```diff
@@ -133,6 +223,46 @@
           or "Registered" in __rhc_insights_status.stdout
       changed_when: true
 
+- name: Configure file redaction
+  when: (rhc_insights.file_redaction.commands is defined) or
```
richm, line 227: Does there need to be a check first for `rhc_insights.file_redaction is defined`?

richm:
How can we test this?

tasks/insights-client.yml
Tronde: AFAICS `rhc_insights.file_redaction` is defined in `defaults/main.yml`. As `defaults/main.yml` as well as `tasks/insights-client.yml` are part of one role and shipped together I do not see a scenario where `rhc_insights.file_redaction` would not be defined. So with my current knowledge and understanding I would say there is no need to check first for `rhc_insights.file_redaction is defined`.

I see that most if not all tasks in `tasks/insights-client.yml` check whether variables are defined although all of these variables are defined in `defaults/main.yml`. TBH I do not understand why these checks are necessary. Are they in place for "just in case" scenarios or is there some reason I do not see?

Following my own argument I would suggest removing all the `var is defined` checks where the var is defined in `defaults/main.yml` as the check is not striclty necessary. But I might miss a good reason here.

templates/file-content-redaction.yaml.j2
```diff
@@ -0,0 +1,43 @@
...
+#     - abc.*
+#     - localhost[[:digit:]]
+#
+# Lines matching these regular expressions will be omitted
+# from output.
+# NOTE: You cannot mix plain string matching and regular expression matching.
+{% if rhc_insights.file_content_redaction.keywords %}
+keywords:
+{% for keyword in rhc_insights.file_content_redaction.keywords %}
+  - {{ keyword }}
```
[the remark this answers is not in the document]
Tronde: For patterns (with regex) that would be possible as the examples in [YAML-style denylist configuration for Red Hat Insights Client](https://access.redhat.com/articles/4511681) show characters that need to be escaped.

For keywords, I tend to escape them just to be sure. Do you agree?

tasks/insights-client.yml
richm: > AFAICS `rhc_insights.file_redaction` is defined in `defaults/main.yml`.

That unfortunately doesn't mean anything.

> As `defaults/main.yml` as well as `tasks/insights-client.yml` are part of one role and shipped together I do not see a scenario where `rhc_insights.file_redaction` would not be defined.

What if I run the role like this?
```
inventory:....
  rhc_insights:
    foo: bar
```
Defining a host variable takes precedence over settings in defaults/main.yml.  The same applies to setting the variable at the play level or as one of the `vars` in the `include_role` call.  Setting `rhc_insights` in this way *completely replaces the existing contents defined in defaults/main.yml*.  https://docs.ansible.com/ansible/latest/reference_appendices/general_precedence.html#variables

So - while "top level" variables like `rhc_insights` will always be defined, their "sub-parameters" may or may not be defined - that is - `rhc_insights is defined` will always return `true`, but `rhc_insights.file_redaction is defined` can return `true` or `false`.

If you don't believe me, try running the role with different values and see what happens.

> So with my current knowledge and understanding I would say there is no need to check first for `rhc_insights.file_redaction is defined`.
> 
> I see that most if not all tasks in `tasks/insights-client.yml` check whether variables are defined although all of these variables are defined in `defaults/main.yml`. TBH I do not understand why these checks are necessary. Are they in place for "just in case" scenarios or is there some reason I do not see?

Yes, and if some checks are missing, they should be added in another PR.

TBH, I'm really not sure why "sub-parameter" keys and values are defined in `defaults/main.yml`, except possibly as a convenience when the user does not provide a value for `rhc_insights` at all.  There is no mechanism in Ansible like "merge the default dict valued parameter `rhc_insights` defined in `defaults/main.yml` with the `rhc_insights` parameter given by the user".  If the user defines `rhc_insights`, then the value in `defaults/main.yml` is completely replaced.

> 
> Following my own argument I would suggest removing all the `var is defined` checks where the var is defined in `defaults/main.yml` as the check is not striclty necessary. But I might miss a good reason here.

templates/file-content-redaction.yaml.j2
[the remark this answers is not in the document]
richm: This is tricky - afaik there is no "regex quote" or "regex escape" filter in Jinja or Ansible (like there is for shell metacharacters and the `quote` filter).  If you use `- "{{ pattern }}"` I think this will cause problems if `pattern` contains `"` - same with `'{{ pattern }}'` and `'` - however, this might be safe:
```yaml
  - >-
   {{ pattern }}
```

ptoscano:
Sorry for the late answer on this.

While I appreciate the efforts on this, I have to raise my concerns on this implementation. The general "style" of the `rhc` role is to work on use cases & functionalities related to the connection of a system to Red Hat / Satellite, and functionalities related to maintaining such. This means for example ensuring a system is connected & registered, it can get content, and there are certain Insights features available. Because of this, the role does not expose configuration bits of the underlying tools simply because they exist, and only in case they are needed, sometimes with different semantics than the underlying tools (e.g. `rhc_insights.tags`), or even that do not map at all (e.g. `rhc_repositories`, `rhc_insights.remediation`).

What I'm missing here are the actual use cases & stories behind the addition of those options. Sometimes configuring an option may be the right solution, and sometimes a problem can be solved in a different way. Unfortunately, without knowing them, it is hard to determine whether adding the options is the correct possibility, and even testing whether the use cases were actually solved. Also, new options mean additional maintenance cost, which thus needs to be "justified".

In addition to that, I find some options problematic to add:
- `rhc_insights_auth`: the authentication was explicitly designed to be the same as used for registering a system, as a separate authentication for Insights is an extreme corner case; in addition to that, basic authentication for Insight is officially deprecated, and will be disabled at the end of this year (2024) [1]
- `rhc_insights.autoconfig` & `rhc_insights.baseurl`: these two option are used to manually configure where `insights-client` connects to, without detecting that from the way the system is registered with `subscription-manager`; this is yet another corner case, mostly to support `insights-client` registered using basic authentication (and thus not able to automatically detect from `subscription-manager`)
- `rhc_insights.logevel`: the default behaviour of `insights-client` is to even log debug messages to its log file (`/var/log/insights-client/insights-client.log`); it is not clear to me why this needs to be configurable via the `rhc` role

Regarding the other two "blocks" of options, i.e. redaction and obfuscation: these big features definitely require use cases and discussions on what is expected, what is needed, and so forth, and what would be the proper API to support them.

Because of what I said, my recommendation for this PR would be to hold it, and move the discussion as internal with us, and determine the various use cases, etc. 

[1] https://docs.redhat.com/en/documentation/red_hat_insights/1-latest/html/release_notes/april-2024#remote_host_configuration_rhc_and_the_insights_client

PR #189, opened 2024-07-30
fix: drop usage of "auto_attach" of the "redhat_subscription" module

ptoscano:
`auto_attach` is supposed to automatically attach subscriptions for accounts in entitlement mode. in Simple Content Access mode, this option becomes a no-op in subscription-manager.

The usage of this option was originally added to help testing regardless of the content access mode, even if the role supports only SCA. Since the entitlement mode has been officially deprecated for a couple of years already, and it is going away soon, then stop using the `auto_attach` option of the `redhat_subscription` module.

There is no behaviour change.

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -40,9 +40,6 @@
     server_prefix: "{{ rhc_server.prefix | d(omit) }}"
     server_insecure: "{{ omit if (rhc_server.insecure | d(omit) == omit)
       else rhc_server.insecure | ternary('1', '0') }}"
-    auto_attach: "{{ true
-      if (rhc_state | d('present') in ['present', 'reconnect']
-          and not 'keys' in rhc_auth.activation_keys | d({})) else omit }}"
     release: "{{ rhc_release
       if (rhc_state | d('present') != 'absent'
            and rhc_release != __rhc_state_absent


ptoscano:
[citest]

ptoscano:
[citest bad]

spetrosi:
Won't this break EL 7, 8, 9?

spetrosi:
Please rebase

ptoscano:
> Won't this break EL 7, 8, 9?

Nope; SCA works fine in those versions.

ptoscano:
[citest]

richm (approved):

ptoscano:
hmmm `tests_repositories` fails now:
```
This system has no repositories available through subscriptions
```
I guess I need to tweak the test data - will need to tweak the test setup a bit.

richm:
> hmmm `tests_repositories` fails now:
> 
> ```
> This system has no repositories available through subscriptions
> ```
> 
> I guess I need to tweak the test data - will need to tweak the test setup a bit.

ok - but it is not related to this PR

ptoscano:
> > hmmm `tests_repositories` fails now:
> > ```
> > This system has no repositories available through subscriptions
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I guess I need to tweak the test data - will need to tweak the test setup a bit.
> 
> ok - but it is not related to this PR

True, however the result is that the tests are broken :/

I should have hopefully fixed my candlepin-unofficial container image, and local changes to switch the testing account seems to work fine.

I'll need #191 to fix one issue in the test setup, and then I can switch the account to a SCA one.

ptoscano:
> I'll need #191 to fix one issue in the test setup, and then I can switch the account to a SCA one.

... which is #195, hopefully.

ptoscano:
[citest]

ptoscano:
Now it should be fine :-)

spetrosi (approved):

PR #191, opened 2024-08-01
tests: factorize & fix URL for candlepin owner

ptoscano:
The Candlepin URL used to query the owner (= organization) used in the tests when using the self-deployed Candlepin was wrong, using the username rather than the organization name. It works fine because the username and the organization name are currently the same.

Hence, to avoid issues in the future, properly use the organization name instead. To simplify the maintenance of that code, create a new helper fact with the Candlepin URL for the current owner.

There is no behaviour change in the tests.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -18,6 +18,10 @@
       set_fact:
         _cp_url: "https://{{ lsr_rhc_test_data.candlepin_host }}:{{ lsr_rhc_test_data.candlepin_port }}{{ lsr_rhc_test_data.candlepin_prefix }}"  # yamllint disable-line
 
+    - name: Set helper fact for Candlepin owner URL
+      set_fact:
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

ptoscano:
[citest bad]

richm (approved):

ptoscano:
[citest bad]

richm:
[citest bad]

richm:
[citest_bad]

richm:
[citest]

ptoscano:
[citest]

spetrosi:
[citest_bad]

richm:
I think we should go ahead and merge this PR.  The failures are due to known issues not related to this PR, and we have a plan to address them.

ptoscano:
> I think we should go ahead and merge this PR. The failures are due to known issues not related to this PR, and we have a plan to address them.

OK. I simply want to avoid forgetting about the issues, and then CI jobs keeps becoming more and more red :-(

PR #195, opened 2024-08-05
tests: switch to an SCA account for Candlepin

ptoscano:
The role officially supports only SCA, so switch the test account used in the self-deployed Candlepin to an SCA-only account.

As consequence, also adapt the product certificate & repositories used to what the new account can access to.

There are no behaviour changes.

### DIFF
--- a/tests/files/candlepin_data.yml
+++ b/tests/files/candlepin_data.yml
@@ -5,16 +5,16 @@ lsr_rhc_test_data:
   candlepin_port: 8443
   candlepin_prefix: /candlepin
   candlepin_insecure: false
-  reg_username: "admin"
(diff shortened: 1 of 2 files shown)


ptoscano:
[citest]

richm (approved):

ptoscano:
[citest]

ptoscano:
Since:
- all the tests now pass, with the exception of the `CentOS-7-latest/ansible-2.17` & `CentOS-Stream-8/ansible-2.17` BaseOS CI jobs that represent untestable combinations
- Rich already approved this
- there was no change since the last review, only rebases

hence I'm merging this.

PR #196, opened 2024-08-06
tests: run podman directly instead of using containers.podman

ptoscano:
Using the `containers.podman` collection to handle `podman` is a nice idea, however what that collection supports now is becoming more strict than what this role supports. Case in point: `containers.podman` now requires Python 3.6+ for the managed nodes, which breaks the testing of EL 7.

This means that sadly the setup of a self-deployed Candlepin needs to run `podman` commands manually; use the proper parameters for `podman run` depending on the OS.

The gymnastics of checking the return code & stderr of `podman stop` are needed because the `--ignore` option (which would be ideal) was introduced in `podman` 1.7.0, and EL 7 has `podman` 1.6.4.

There is no behaviour change.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -37,23 +37,41 @@
                  ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
     - name: Stop and remove Candlepin container
-      containers.podman.podman_container:
-        name: candlepin
-        state: absent
+      command:
+        argv:
+          - podman
+          - stop
+          - candlepin
+      register: podman_stop_status
+      failed_when:
+        - podman_stop_status.rc != 0
+        - '"no such container" not in podman_stop_status.stderr'
+      changed_when: false
 
     - name: Start Candlepin container
-      containers.podman.podman_container:
-        detach: true
-        hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
-        image: ghcr.io/ptoscano/candlepin-unofficial
-        name: candlepin
-        privileged: "{{ ansible_distribution in ['CentOS', 'RedHat']
-          and ansible_distribution_major_version | int < 8 }}"
-        publish:
+      vars:
+        podman_run_cli:
+          - podman
+          - run
+          - --rm
+          - --detach
+          - --hostname
+          - "{{ lsr_rhc_test_data.candlepin_host }}"
+          - --name
+          - candlepin
+          - --publish
           - 8443:8443
+          - --publish
           - 8080:8080
-        rm: true
-        state: started
+          - "{{ '--privileged'
+            if (ansible_distribution in ['CentOS', 'RedHat']
+                and ansible_distribution_major_version | int < 8)
+            else '' }}"
+          - ghcr.io/ptoscano/candlepin-unofficial
+      command:
+        argv:
+          "{{ podman_run_cli | reject('match', '^$') | list }}"
+      changed_when: false
 
     - name: Ensure directories exist
       file:
--- a/tests/collection-requirements.yml
+++ b/tests/collection-requirements.yml
@@ -1,4 +1,3 @@
 ---
 collections:
-  - name: containers.podman
   - name: fedora.linux_system_roles


ptoscano:
[citest]

tests/tasks/setup_candlepin.yml
```diff
@@ -37,23 +64,22 @@
...
+          - podman
+          - stop
+          - candlepin
+      register: podman_stop_status
+      failed_when:
+        - podman_stop_status.rc != 0
+        - '"no such container" not in podman_stop_status.stderr'
+      changed_when: false
 
     - name: Start Candlepin container
```
spetrosi, line 78:
```suggestion
    - name: Start Candlepin 
      vars:
        podman_run_cli:
          - podman
          - run
          - --rm
          - --detach
          - --hostname
          - "{{ lsr_rhc_test_data.candlepin_host }}"
          - --name
          - candlepin
          - --publish
          - 8443:8443
          - --publish
          - 8080:8080
          - "{{
            '--privileged'
            if ansible_distribution in ['CentOS', 'RedHat'] and ansible_distribution_major_version | int < 8
            else '' }}"
          - ghcr.io/ptoscano/candlepin-unofficial    
```

Instead of setting three facts, let's set the cli here

ptoscano: oh wow, is this syntax actually working? let's see...

ptoscano: this doesn't work: in case `--priviledged` is not set, the result is an empty item in the list, and this empty item becomes an empty parameter to podman, which then fails:
```
STDERR:

Error: parsing reference "": repository name must have at least one component

```

spetrosi: Oh, okay. One more thing you can try is using `omit`, i.e. replace `else '' }}"` with `else omit }}"`.
And if this doesn't work too, you can do what you did with set_fact but with `vars`, like this:

```
podman_run_cli: "{{
  podman_run_cli + ['--privileged']
  if ansible_distribution in ['CentOS', 'RedHat'] and ansible_distribution_major_version | int < 8
  else podman_run_cli }}"
podman_run_cli: "{{ podman_run_cli + ['ghcr.io/ptoscano/candlepin-unofficial'] }}"
```

ptoscano: ah, a bit of jinja filtering will do the job -- thanks for the hint!

spetrosi: set_fact is not welcomed because it creates an Ansible fact that is kept after playbook finished, we want to avoid this if we can.

ptoscano:
[citest]

ptoscano:
[citest]

ptoscano:
[citest]

tests/tasks/setup_candlepin.yml
```diff
@@ -22,6 +22,31 @@
...
+          - 8080:8080
+          - "{{ '--privileged'
+            if (ansible_distribution in ['CentOS', 'RedHat']
+                and ansible_distribution_major_version | int < 8)
+            else '' }}"
+          - ghcr.io/ptoscano/candlepin-unofficial
+
+    - name: Filter empty items from podman run invocation
+      set_fact:
+        podman_run_cli: "{{ podman_run_cli | reject('match', '^$') | list }}"
```
spetrosi, line 48: That's cool, if it works you can still pass both set_facts as `vars` in the `Start Candlepin container` to avoid setting facts.

spetrosi: Why won't you do vars like this? From what I see in the PR, podman_run_cli is not used anywhere else.
```
      vars:
        podman_run_cli:
          - podman
          - run
          - --rm
          - --detach
          - --hostname
          - "{{ lsr_rhc_test_data.candlepin_host }}"
          - --name
          - candlepin
          - --publish
          - 8443:8443
          - --publish
          - 8080:8080
          - "{{ '--privileged'
            if (ansible_distribution in ['CentOS', 'RedHat']
                and ansible_distribution_major_version | int < 8)
            else '' }}"
          - ghcr.io/ptoscano/candlepin-unofficial
        podman_run_cli: "{{ podman_run_cli | reject('match', '^$') | list }}"
```

spetrosi: Sorry, ansible doesn't like that we use variables with the same name. This works for me:
```
- name: debug
  vars:
    podman_run_cli_full:
      - podman
      - run
      - --rm
      - --detach
      - --hostname
      - "{{ lsr_rhc_test_data.candlepin_host }}"
      - --name
      - candlepin
      - --publish
      - 8443:8443
      - --publish
      - 8080:8080
      - "{{ '--privileged'
        if (ansible_distribution in ['CentOS', 'RedHat']
        and ansible_distribution_major_version | int < 8)
        else '' }}"
      - ghcr.io/ptoscano/candlepin-unofficial
    podman_run_cli: "{{ podman_run_cli_full | reject('match', '^$') | list }}"
  debug:
    var: podman_run_cli

```

ptoscano: I tried that too, however it doesn't work either:
```
ok: [localhost] => {
    "podman_run_cli": "VARIABLE IS NOT DEFINED!"
}
```
I'm using ansible-core 2.16.9.

spetrosi: How about this:
```
- name: debug
  vars:
    podman_run_cli_full:
      - podman
      - run
      - --rm
      - --detach
      - --hostname
      - candlepin_host
      - --name
      - candlepin
      - --publish
      - 8443:8443
      - --publish
      - 8080:8080
      - "{{ '--privileged'
        if (ansible_distribution in ['CentOS', 'RedHat']
        and ansible_distribution_major_version | int < 8)
        else '' }}"
      - ghcr.io/ptoscano/candlepin-unofficial
    podman_run_cli: "{{ podman_run_cli_full | reject('match', '^$') | list }}"
  debug:
    var: podman_run_cli
```

Previous didn't work because lsr_rhc_test_data is not defined. But will work in the context of the role.

richm:
Another way to dynamically construct lists in an assignment statement where you want to omit some of the elements is like this:

```yaml
    - name: Start Candlepin container
      command:
        argv:
          "{{ podman_run_cli }}"
      changed_when: false
      vars:
        podman_run_cli: [podman, run, --rm, --detach, --hostname, candlepin_host, --name, candlepin, --publish,
                                        8443:8443, --publish, 8080:8080] +
                                        {{ ['--privileged'] if (ansible_distribution in ['CentOS', 'RedHat']
                                            and ansible_distribution_major_version | int < 8)
                                            else [] }} + [ghcr.io/ptoscano/candlepin-unofficial]
```

Yes, it uses the "json" style of list definition, but I prefer readability to strict adherence of Ansible guidelines

tests/tasks/setup_candlepin.yml
ptoscano: thanks for the hints! I got a working version of it, simplified a little bit

ptoscano:
[citest]

spetrosi (approved):

tests/tasks/setup_candlepin.yml
```diff
@@ -37,23 +37,41 @@
...
           - 8443:8443
+          - --publish
           - 8080:8080
-        rm: true
-        state: started
+          - "{{ '--privileged'
+            if (ansible_distribution in ['CentOS', 'RedHat']
+                and ansible_distribution_major_version | int < 8)
+            else '' }}"
+          - ghcr.io/ptoscano/candlepin-unofficial
```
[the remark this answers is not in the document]
richm: And yes - it is unfortunate that ansible/jinja/yaml makes this such an awkward construction

richm (approved):

PR #209, opened 2025-01-20
fix: use the right systemd service for remediations in EL 10+

ptoscano:
EL 10 has a newer yggdrasil service for remediations, so instead of 'rhcd' the upstream name 'yggdrasil' is used.

Hence:
- create an helper variable for the service name depending on the EL version; since Insights is specific to RHEL, it is possible to assume the name based on the major version of the distribution
- use the new variable when enabling or disabling the service for remediations
- in the test case for remediation:
  - export the variables of the role as public, so the new private variable can be used
  - set an helper fact with the systemd service including the '.service' extension

Fixes: CCT-1057


tests/tests_insights_remediation.yml
```diff
@@ -41,14 +42,18 @@
...
+
         - name: Get service_facts
           service_facts:
 
         - name: Check remediation is enabled
           assert:
             that:
-              - "'rhcd.service' in ansible_facts.services"
-              - ansible_facts.services['rhcd.service'].status == 'enabled'
+              - "yggdrasil_systemd_service in ansible_facts.services"
```
richm, line 55:
```suggestion
              - yggdrasil_systemd_service in ansible_facts.services
```
This shouldn't be quoted anymore since there are no quoted strings being used anymore

ptoscano: ooh nice, thanks

richm:
[citest]

subpop:
This LGTM. Thanks @ptoscano!

richm (approved):
el10 podman issues are known and unrelated

PR #210, opened 2025-01-20
tests: stop adding pools to the activation key

ptoscano:
The Candlepin account used for tests uses SCA, and thus any activation key for it can already access to the whole content of the organization of the account. The addition of the pools to the activation key is thus redundant, and upcoming versions of Candlepin will not allow this behaviour anymore.

Hence, stop adding pools to the activation key, which is no more needed.

There is no behaviour change.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -135,31 +135,6 @@
         dest: /etc/pki/rpm-gpg/RPM-GPG-KEY-candlepin
         mode: "0644"
 
-    - name: Query for default_key activation key
-      uri:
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

richm (approved):

ptoscano:
[citest bad]

richm:
> The c10s failures are due to `podman` failing to execute the Candlepin container:
> 
> ```
> Error: OCI runtime error: crun: `/proc/self/attr/keycreate`: OCI runtime error: unable to process security attribute
> ```
> 
> The other combinations are OK; should I merge this?

Yes

PR #216, opened 2025-02-04
Adding rhc_pool_ids

fao89:
Enhancement:
Adding pool_ids

Reason:
Enable users to attach a specific pool

Ref:
https://docs.ansible.com/ansible/latest/collections/community/general/redhat_subscription_module.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,14 @@ it was already connected; because of this, the role will always report a
 "changed" status.
 
 ```yaml
+    rhc_pool_ids:
+      - 0123456789abcdef0123456789abcdef: 2
+      - 1123456789abcdef0123456789abcdef: 4
+```
+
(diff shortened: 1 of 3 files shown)


ptoscano:
Hi @fao89,

the `rhc` role was designed with Simple Content Access (SCA) as only content access mode for subscriptions; pools are specific to the other content access mode, i.e. entitlements.

Entitlements have been long deprecated (more than 2 years ago), they will not be supported in upcoming versions of Satellite, and the vast majority of the organizations/accounts directly connected to Red Hat were migrated to SCA a couple of months ago. Please check this extensive documentation:
https://access.redhat.com/articles/transition_of_subscription_services_to_the_hybrid_cloud_console

What is the use case for supporting pools in the role?

PR #220, opened 2025-02-24
tests: adjust UUID checks to newer versions of insights-core

ptoscano:
Starting from version 3.4.6 [1], insights-core (used internally by `insights-client`) adjusted its logic to generate the UUID (written in the `/etc/insights-client/machine-id` file): instead of a new random UUID, it uses the `subscription-manager` UUID in case the system is registered with it (which now is the only supported option).

The expectations in tests_insights_client_register were that, unregistering and registering again with `insights-client` would give a new UUID, which are no more true now. Thus adjust the checks to either check for the old behaviour, or check that the UUID is now the same as the rhsm UUID.

This is only a fix for the test, the behaviour of the role itself is correct.

[1] https://github.com/RedHatInsights/insights-core/pull/4057

### DIFF
--- a/tests/tests_insights_client_register.yml
+++ b/tests/tests_insights_client_register.yml
@@ -74,6 +74,9 @@
             rhc_state: reconnect
             rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
 
+        - name: Get rhsm UUID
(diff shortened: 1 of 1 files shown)


ptoscano:
[citest]

spetrosi (approved):

PR #224, opened 2025-04-02
tests: Cleanup candlepin container

spetrosi:
Enhancement: In always block, import a task to cleanup candlepin container in tests that import the task tests/tasks/setup_candlepin.yml, which starts candlepin container.

Reason: Candlepin container remains after test playbook finishes. If you remove `container-selinux` rpm, you got selinux denials because the container is running but selinux policies for containers are removed.

### DIFF
--- a/tests/tasks/setup_candlepin.yml
+++ b/tests/tasks/setup_candlepin.yml
@@ -36,17 +36,8 @@
         use: "{{ __rhc_is_ostree |
                  ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
-    - name: Stop and remove Candlepin container
-      command:
-        argv:
-          - podman
-          - stop
-          - candlepin
-      register: podman_stop_status
-      failed_when:
-        - podman_stop_status.rc != 0
-        - '"no such container" not in podman_stop_status.stderr'
-      changed_when: false
+    - name: Clean up Candlepin container
+      include_tasks: teardown_candlepin.yml
 
     - name: Start Candlepin container
       vars:
--- a/tests/tests_environments.yml
+++ b/tests/tests_environments.yml
@@ -7,101 +7,106 @@
   tags:
     - tests::slow
   tasks:
-    - name: Setup Candlepin
-      import_tasks: tasks/setup_candlepin.yml
-      vars:
-        environments: true
-
-    - name: Skip if no test environments are set
-      meta: end_play
-      when:
-        - >-
-          lsr_rhc_test_data.envs_register is none
-          or lsr_rhc_test_data.envs_register | d([]) | length == 0
-
-    - name: Ensure ansible_facts used by the test
-      setup:
-        gather_subset:
-          - "!all"
-          - "!min"
-          - distribution
-          - distribution_major_version
-          - distribution_version
-
-    - name: Try to register (wrong environment)
+    - name: Basic repository enablement/disablement test
       block:
-        - name: Register (wrong environment)
-          include_role:
-            name: linux-system-roles.rhc
+        - name: Setup Candlepin
+          import_tasks: tasks/setup_candlepin.yml
           vars:
-            rhc_auth:
-              login:
-                username: "{{ lsr_rhc_test_data.reg_username }}"
-                password: "{{ lsr_rhc_test_data.reg_password }}"
-            rhc_insights:
-              state: absent
-            rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
-            rhc_server:
-              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
-              port: "{{ lsr_rhc_test_data.candlepin_port }}"
-              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
-              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
-            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
-            rhc_environments:
-              - "{{ lsr_rhc_test_data.env_nonworking }}"
+            environments: true
 
-        - name: Unreachable task
-          fail:
-            msg: The above task must fail
-      rescue:
-        - name: Assert registration failed
-          assert:
-            that: ansible_failed_result.msg != 'The above task must fail'
-
-    - name: Test environments on registration
-      block:
-        - name: Register (with existing environments)
-          include_role:
-            name: linux-system-roles.rhc
-          vars:
-            rhc_auth:
-              login:
-                username: "{{ lsr_rhc_test_data.reg_username }}"
-                password: "{{ lsr_rhc_test_data.reg_password }}"
-            rhc_insights:
-              state: absent
-            rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
-            rhc_server:
-              hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
-              port: "{{ lsr_rhc_test_data.candlepin_port }}"
-              prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
-              insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
-            rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
-            rhc_environments: "{{ lsr_rhc_test_data.envs_register }}"
-
-        # 'subscription-manager environments' has a '--list' option only
-        # in RHEL 8.6+ and greater (incl. RHEL 9+)
-        - name: Check the enabled environments
+        - name: Skip if no test environments are set
+          meta: end_play
           when:
             - >-
-              ansible_distribution not in ["CentOS", "RedHat"]
-              or (ansible_distribution == "CentOS"
-                  and ansible_distribution_major_version | int >= 8)
-              or (ansible_distribution_version is version("8.6", ">="))
+              lsr_rhc_test_data.envs_register is none
+              or lsr_rhc_test_data.envs_register | d([]) | length == 0
+
+        - name: Ensure ansible_facts used by the test
+          setup:
+            gather_subset:
+              - "!all"
+              - "!min"
+              - distribution
+              - distribution_major_version
+              - distribution_version
+
+        - name: Try to register (wrong environment)
           block:
-            - name: Get enabled environments
-              include_tasks: tasks/list_environments.yml
+            - name: Register (wrong environment)
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_auth:
+                  login:
+                    username: "{{ lsr_rhc_test_data.reg_username }}"
+                    password: "{{ lsr_rhc_test_data.reg_password }}"
+                rhc_insights:
+                  state: absent
+                rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+                rhc_server:
+                  hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+                  port: "{{ lsr_rhc_test_data.candlepin_port }}"
+                  prefix: "{{ lsr_rhc_test_data.candlepin_prefix }}"
+                  insecure: "{{ lsr_rhc_test_data.candlepin_insecure }}"
+                rhc_baseurl: "{{ lsr_rhc_test_data.baseurl | d(omit) }}"
+                rhc_environments:
+                  - "{{ lsr_rhc_test_data.env_nonworking }}"
 
-            - name: Check environments to enable are enabled
+            - name: Unreachable task
+              fail:
+                msg: The above task must fail
+          rescue:
+            - name: Assert registration failed
               assert:
-                that:
-                  - >-
-                    (lsr_rhc_test_data.envs_register | sort) ==
-                    (test_environments.stdout_lines | sort)
+                that: ansible_failed_result.msg != 'The above task must fail'
 
+        - name: Test environments on registration
+          block:
+            - name: Register (with existing environments)
+              include_role:
+                name: linux-system-roles.rhc
+              vars:
+                rhc_auth:
+                  login:
+                    username: "{{ lsr_rhc_test_data.reg_username }}"
+                    password: "{{ lsr_rhc_test_data.reg_password }}"
+                rhc_insights:
+                  state: absent
+                rhc_organization: "{{ lsr_rhc_test_data.reg_organization }}"
+                rhc_server:
+                  hostname: "{{ lsr_rhc_test_data.candlepin_host }}"
+                  port: "{{ lsr_rhc_test_data.candlepin_port }}"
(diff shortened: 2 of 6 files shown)


spetrosi:
[citest]

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,21 @@
...
+  register: __rhc_candlepin_container
+  changed_when: false
+
+- name: Stop Candlepin container
+  command: podman stop candlepin
+  changed_when: true
+  when: '"candlepin" in __rhc_candlepin_container.stdout'
+
+- name: Remove Candlepin container
+  command: podman rm candlepin
```
richm, line 14: This gives an error because the container was run with the `--rm` argument, which automatically removes the container after `podman stop`
https://github.com/linux-system-roles/rhc/blob/main/tests/tasks/setup_candlepin.yml#L56

I still think it is a good idea to do the `rm` just to ensure that it is gone - maybe do `podman rm --force --ignore candlepin`

spetrosi:
[citest]

spetrosi:
[citest]

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,21 @@
...
+  register: __rhc_candlepin_container
+  changed_when: false
+
+- name: Stop Candlepin container
+  command: podman stop candlepin
+  changed_when: true
+  when: '"candlepin" in __rhc_candlepin_container.stdout'
+
+- name: Remove Candlepin container
+  command: podman rm --force --ignore candlepin
```
richm, line 14: no `--ignore` flag on the podman version in rhel 7 :-(
```
Error: unknown flag: --ignore
```
so remove `--ignore` and add `ignore_errors: true`

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
@ptoscano please take a look when you have time

tests/tasks/cleanup_candlepin_container.yml
ptoscano, line 1: I'd rename this file as "stop_candlepin.yml"

ptoscano, line 1: The added tasks in this new file must be executed only when a Candlepin container is configured to be used (i.e. the default, and when an external configuration is not set).
Follow the same logic done at the top of `setup_candlepin.yml`, i.e.
```yaml
- name: Stop Candlepin
  when: lsr_rhc_test_data_file | length == 0
  block:
    # all the new tasks here
```
This way this stop task will not stop an existing/running "candlepin" container when not configured to use it.

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,37 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Check if the candlepin container is started
+  command: podman container exists candlepin
```
ptoscano, line 4: please use the `argv:` syntax for `command`, ie
```suggestion
  command:
    argv:
      - podman
      - container
      - exists
      - candlepin
```
While it is more verbose, it is consistent with what already used everywhere in the role, and it has the big advantage of not requiring shell split by the `command` module.

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,37 @@
...
+  register: __rhc_candlepin_cont_started
+  failed_when: false
+  changed_when: false
+
+- name: Stop and potentionally remove Candlepin container
+  command: podman stop candlepin
```
ptoscano, line 10: ditto

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,37 @@
...
+  when: __rhc_candlepin_cont_started.rc == 0
+  changed_when: true
+
+- name: Check if there is a stopped candlepin container
+  command: podman ps -a --filter name=candlepin
```
ptoscano, line 15: ditto

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,37 @@
...
+  register: __rhc_candlepin_container_stopped
+  failed_when: false
+  changed_when: false
+
+- name: Remove Candlepin container
+  command: podman rm candlepin
```
ptoscano, line 21: ditto

tests/tasks/cleanup_candlepin_container.yml
ptoscano: Well, if I'm running podman with `--rm`, then I expect the container to be gone on `stop`, and not require and removal manually... If it does not do that, I'd consider it a podman bug.

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,15 @@
...
+- name: Stop Candlepin container
+  command: podman stop candlepin
+  changed_when: true
+  when: '"candlepin" in __rhc_candlepin_container.stdout'
+
+- name: Remove Candlepin container
+  command: podman rm candlepin
+  register: podman_stop_status
+  changed_when: true
+  when: '"candlepin" in __rhc_candlepin_container.stdout'
```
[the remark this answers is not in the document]
ptoscano: Is removing the image from the local registry actually needed?
- in ephemeral test machines this operation is practically of no use (the machine will be dropped/reprovisioned/etc)
- in persistent systems the existing image is reused when running another test, and at worst a new image will be downloaded

While I understand the point of "cleaning up", in this case IMHO this additional logic does not give us a lot of benefits. Even worse, it will cause to download the image every time a test is run in persistent systems, in case it did not change from the registry.

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,37 @@
...
+- name: Check if the candlepin container is started
+  command: podman container exists candlepin
+  register: __rhc_candlepin_cont_started
+  failed_when: false
+  changed_when: false
+
+- name: Stop and potentionally remove Candlepin container
+  command: podman stop candlepin
+  when: __rhc_candlepin_cont_started.rc == 0
+  changed_when: true
```
ptoscano, line 12: Instead of these two tasks, simply use the existing `Stop and remove Candlepin container` task in `setup_candlepin.yml`, as it does exactly that.

ptoscano (changes requested):

richm:
@ptoscano re: removing the candlepin image from the local registry

* tests should clean up all resources - you are correct that if this is an ephemeral machine, none of the cleanup steps are necessary - what we could do for performance is tag all of those tasks with `tests::cleanup`, then use `--skip-tags tests::cleanup` if the machine is ephermeral - but we definitely need the cleanup for the default case, which is non-ephemeral
* leaving the container image may cause problems in multi-role tests e.g. the podman tests expect a system without any local containers or images

ptoscano:
>  what we could do for performance is tag all of those tasks with `tests::cleanup`, then use `--skip-tags tests::cleanup` if the machine is ephermeral - but we definitely need the cleanup for the default case, which is non-ephemeral

I'm fine with this. Could we also tweak tox-lsr (which spawns a new system) to also skip that tag by default?

> The problem is that we don't have a way to run a "cleanup-after-all-role-tests-are-completed.yml" playbook.

I guess it also depends on how the tests are run, right? ie tox-lsr vs testing-farm vs something else.

richm:
> > what we could do for performance is tag all of those tasks with `tests::cleanup`, then use `--skip-tags tests::cleanup` if the machine is ephermeral - but we definitely need the cleanup for the default case, which is non-ephemeral
> 
> I'm fine with this. Could we also tweak tox-lsr (which spawns a new system) to also skip that tag by default?

It's pretty easy to add `--skip-tags tests::cleanup` to your tox qemu run invocation e.g. `tox -e qemu-ansible-core-2.17 -- --image-name centos-10 --log-level debug --skip-tags tests::cleanup -- tests/tests_repositories.yml`

There are other ways to use tox e.g. with `--make-batch` where you do not want to skip cleanup.

> 
> > The problem is that we don't have a way to run a "cleanup-after-all-role-tests-are-completed.yml" playbook.
> 
> I guess it also depends on how the tests are run, right? ie tox-lsr vs testing-farm vs something else.

Right - if we wanted to add something like a "run setup before all role tests" and "run cleanup after all role tests" then we would have to change tox-lsr, testing farm test runner, and the downstream test runners.

tests/tasks/cleanup_candlepin_container.yml
```diff
@@ -0,0 +1,37 @@
...
+  failed_when: false
+  changed_when: false
+
+- name: Remove Candlepin container
+  command: podman rm candlepin
+  changed_when: true
+  when:
+    - '"candlepin" in __rhc_candlepin_container_stopped.stdout'
+
+- name: Remove Candlepin container image
```
spetrosi, line 26: @richm maybe let's not clean up the image? It won't affect anything, and re-downloading it with each test upstream is a waste of network bandwidth.

richm: > @richm maybe let's not clean up the image? It won't affect anything, and re-downloading it with each test upstream is a waste of network bandwidth.

Ok.  That will affect the podman tests, in the scenarios where we run tests from multiple roles - the podman tests expect no images.  I guess that means changing this PR to use `__rhc_cleanup_image: false` (or just omit __rhc_cleanup_image because the default is false)

richm: I'll fix the podman tests to work if there are images on the system.

spetrosi:
[citest]

tests/tests_environments.yml
```diff
@@ -7,101 +7,108 @@
...
+                rhc_state: absent
       always:
-        - name: Unregister
-          include_role:
-            name: linux-system-roles.rhc
+        - name: Clean up Candlepin container
+          include_tasks: tasks/remove_candlepin_container.yml
           vars:
-            rhc_state: absent
+            __rhc_cleanup_image: true
```
richm, line 114:
```suggestion
```
I don't think this is needed anymore

richm:
I think just get rid of `__rhc_cleanup_image` which isn't used anymore, and we can merge

tests/tasks/cleanup_candlepin_container.yml
spetrosi: Since in fact we stop the container to remove it, and since this is specifically for the podman container,, I find it clearer to call the file remove_candlepin_container.

spetrosi:
I'd wait for Pino approval too

spetrosi:
[citest]

tests/tasks/remove_candlepin_container.yml
ptoscano, line 1: Since this file now only stops candlepin, better name it `stop_candlepin.yml`, or maybe even better `teardown_candlepin.yml` (symmetric with "setup")

ptoscano, line 1: Rather than two tasks, simply try to stop the container, and do not consider the missing container an error. In practice simply move here the single task `Stop and remove Candlepin container` as currently in `setup_candlepin.yml`.

tests/tasks/remove_candlepin_container.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Stop and remove Candlepin container
+  command:
+    argv:
+      - podman
+      - stop
+      - candlepin
+  changed_when: true
+  when:
+    - lsr_rhc_test_data_file | length == 0
```
ptoscano, line 22: Rather than checking this only for a specific task, better apply a whole `block` for all the tasks, mimicking what `setup_candlepin.yml` does. This way:
- this task file can be imported without checking what's the testing configuration (similar to `setup_candlepin.yml`)
- in case more tasks need to be done for stopping candlepin, they can be added in the existing block

tests/tasks/setup_candlepin.yml
```diff
@@ -36,17 +36,8 @@
...
-          - podman
-          - stop
-          - candlepin
-      register: podman_stop_status
-      failed_when:
-        - podman_stop_status.rc != 0
-        - '"no such container" not in podman_stop_status.stderr'
-      changed_when: false
+    - name: Remove Candlepin container
+      include_tasks: remove_candlepin_container.yml
```
ptoscano, line 40: According to my previous suggestions for the new task file for stopping candlepin, this new task file can be imported outside of this block, between the `Setup the test data` task and `Deploy Candlepin`

ptoscano (changes requested):
Thanks for the updates! I added few more notes that IMHO simplify things a bit, and make the changes more consistent with the current tasks.

Additionally a minor nit common to all the files:
```
        - name: Clean up Candlepin container
          include_tasks: tasks/remove_candlepin_container.yml
```
I suggest `Clean up Candlepin` instead (mimicks the setup tasks), or even `Tear down Candlepin` (see my previous comment on the naming of the new task file).

Thanks!

spetrosi:
[citest]

tests/tasks/remove_candlepin_container.yml
spetrosi: Now this doesn't work because setup_candlepin.yml has the step to remove the container just in case the container was running from the previous playbook. But role always cleans up the container, and this step would fail. Although, it's good to have the step in setup_candlepin.yml for the case when you are testing locally and might have some container running from the previous test.
I think it's better to keep it the Ansible way and keep the task. By importing the task we say bring the system to the state where this container is removed, whether it exists now or not.

richm (approved):

tests/tasks/setup_candlepin.yml
```diff
@@ -3,6 +3,17 @@
...
+  package:
+    name:
+      - podman
+    state: present
+    use: "{{ __rhc_is_ostree |
+              ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
+
+- name: Remove Candlepin container
+  include_tasks: teardown_candlepin.yml
+
```
ptoscano, line 16: Hm there is no need to install podman when running the tests using an external registration server, especially that it takes many seconds.

Apologies for my previous suggestion, let's move this back in the block.

tests/tasks/remove_candlepin_container.yml
ptoscano: Not sure what you mean here: the existing task that tries to stop the candlepin container also takes care of not erroring out in case the container does not exist. Strictly speaking, doing first the existence check and then stopping it may be a minor TOCTOU issue: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use.

ptoscano (changes requested):
Thanks for the updates, a couple of final changes, and it should be ready.

Make sure to squash all the commits.

spetrosi:
[citest]

PR #227, opened 2025-04-17
Force rhc_organization variable to be interger

bshephar:
Enhancement:
All values for `rhc_organization` should be interger values. This change asserts that requirement by ensuring the value is an integer before trying to register.

Reason:
In situations where we're unmarshaling data from json to form variables, it's possible that those values can be unmarshalled as floating point numbers. Which would be incompatible with the current org ID's. This change asserts that the type of number for rhc_organization should be an integer.

Result:

Jira: https://issues.redhat.com/browse/OSPRH-15097

### DIFF
--- a/tasks/subscription-manager.yml
+++ b/tasks/subscription-manager.yml
@@ -31,7 +31,7 @@
     password: "{{ rhc_auth.login.password | d(omit) }}"
     activationkey: "{{ rhc_auth.activation_keys['keys'] | join(',')
       if ('keys' in rhc_auth.activation_keys | d({})) else omit }}"
-    org_id: "{{ rhc_organization
+    org_id: "{{ rhc_organization | string | default(omit)
       if (not rhc_organization is none
           and rhc_organization != omit)
       else omit }}"


ptoscano:
> All values for `rhc_organization` should be interger values.

This is not correct: the organization ID is defined as string in the registration server:
https://github.com/candlepin/candlepin/blob/d519490f437bda6708b7544d298fec124790ea09/api/candlepin-api-spec.yaml#L9662-L9700
(see the "id" property in particular)
It is handled as such in the client tooling (i.e. subscription-manager, rhc), which use both the user input and the API responses as-is.

It looks like an integer now, yes, however there is no requirement for it to be so. It is also documented to be a string both in the README.md of this repository and in the RHEL documentation:
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/automating_system_administration_by_using_rhel_system_roles/using-the-rhc-system-role-to-register-the-system_automating-system-administration-by-using-rhel-system-roles

ptoscano:
> We can `| string` instead to ensure type which would resolve the underlying issue we have.

... but why? Many of the parameters of the role are strings, and I do not see why `rhc_organization` is "more special" than the others. I did read the linked issue, and from my POV it feels a mistake of whatever handles the credentials (that end up also as `rhc_organization`) that handles the org ID as non-string at some point. Especially...

> then we'll just end up with `12345678.0` turning into `"12345678.0"`. So, we need to assert the int for the .0 to be dropped.

... because I find this as clear issue of whatever fills `rhc_organization`. Please fix that instead.

PR #282, opened 2026-05-18
refactor: use ansible.posix 2.1.X for EL7 compatibility [citest_skip]

richm:
The recently released ansible.posix 2.2.0 does not work on EL7.
Pin the version of ansible.posix to 2.1.X.

NOTE: Even though this role might not support EL7, this update
is applied to all system roles for consistency.  Plus, when this
role is part of the system roles collection, all roles must use
the same version of ansible.posix - there is no way for a role
which is part of a collection to use a different version of a
dependency than the version used by the other roles.

### DIFF
--- a/meta/collection-requirements.yml
+++ b/meta/collection-requirements.yml
@@ -2,5 +2,6 @@
 ---
 collections:
   - name: ansible.posix
+    version: '>=2.1.0,<2.2.0'
   - name: community.general
     version: '>=6.6.0,<12.0.0'


meta/collection-requirements.yml
```diff
@@ -2,5 +2,6 @@
 ---
 collections:
   - name: ansible.posix
+    version: 2.1.0
```
ptoscano: Maybe then better with an upper bound instead?
```suggestion
    version: '<2.2'
```

richm: the new one is `version: <2.2.0`

ptoscano (approved):

PR #299, opened 2026-08-20
fix: Select RHSM modules by managed node python version

spetrosi:
Enhancement: Select the RHSM modules by the managed node's python version instead of the ansible version. Gate each of the three subscription-manager tasks (`redhat_subscription`, `rhsm_release`, `rhsm_repository`) on `ansible_facts.python_version`: use the dependency-free vendored `*_python_36` module when python < 3.7, and the bare module name (redirecting to the latest `community.general` module) when python >= 3.7. Rename the vendored modules from `*_ansible_29` to `*_python_36` to reflect the latest python they back, gather the `python_version` fact subset, and drop the collection-requirements version constraints so modern systems install the latest ansible.posix and community.general.

Reason: The latest `community.general` RHSM modules use `from __future__ import annotations`, which requires python >= 3.7 and fails to import on managed nodes running python 3.6, regardless of the controller's ansible version. An ansible-version gate does not prevent this because the crash depends on the managed node's python. The vendored modules import only core `ansible.module_utils`, so they carry no collection dependency and also cover ansible 2.9 hosts (e.g. EL7) where the collection cannot be installed. A fully-qualified name is avoided because an unresolvable FQCN aborts play parsing on ansible 2.9 even for a task skipped by `when`, so the bare name is used.

Result: The role runs the vendored `*_python_36` modules on python 3.6 (and py2.7/EL7) managed nodes, while using the latest `community.general` modules on nodes with python >= 3.7.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added compatibility for managing subscriptions, releases, and repositories on systems running Python versions below 3.7.
  * Subscription management supports registration, unregistration, pool subscriptions, repository configuration, release settings, and system-purpose updates.
  * Added repository selection with wildcard matching, purge support, check mode, and before/after change details.
  * Added distribution detection for Red Hat-based versions 7 through 10.

* **Chores**
  * Removed fixed version constraints for supported Ansible collections.

### DIFF
--- a/library/redhat_subscription_ansible_29.py
+++ b/library/redhat_subscription_ansible_29.py
@@ -12,7 +12,7 @@ __metaclass__ = type
 
 
 DOCUMENTATION = r"""
-module: redhat_subscription_el7
+module: redhat_subscription_ansible_29
 short_description: Manage registration and subscriptions to RHSM using C(subscription-manager)
 description:
   - Manage registration and subscription to the Red Hat Subscription Management entitlement platform using the C(subscription-manager)
--- a/library/redhat_subscription_el7.py
+++ b/library/redhat_subscription_el7.py
@@ -1,0 +1,1295 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) James Laska (jlaska@redhat.com)
+#
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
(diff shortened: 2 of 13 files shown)


spetrosi:
[citest]

spetrosi:
[citest]

richm (approved):

richm (approved):

richm:
[citest]

richm:
[citest_bad]

richm:
[citest_bad]

richm (approved):

